PR #4, opened 2014-11-26
Make it possible to work on something else than servlets

dgageot:
It would be nice to be able to reuse this code in a server that is not based on servlets.
For example, most of the code from `VmRuntimeWebAppContext` could be used on 
a plain http server.

I'm not aiming for a merge right away but more for an open discussion wether this is something
you might be interested in.


ludoch:
All the changes + some extra refactoring from Internal development are in now. Closing this PR, double check this is working for you.

PR #6, opened 2014-12-09
Remove Jetty XML dependency.

alesj:


ludoch:
The code is now there, but via another change than this pull request.

Issue #12, opened 2015-04-08
Simpify XML parsing with XSD and JAXB


jboynes:
AppEngineWebXml is constructed manually by parsing the file to a DOM and walking the tree. No validation is performed.

We could simplify this by define a XSD to describe the document and then using JAXB to parse it into a object tree.

PR #13, opened 2015-04-08
Use JAXB to parse appengine-web.xml

jboynes:
Fix for #12
Rather than parse and process XML manually, this change uses XJC to generate Java objects from the appengine-web.xsd schema definition and JAXB to parse the XML file. It updates the references to the old objects in this project but would be a breaking change for anyone else using that model.


ludoch:
We cannot currently accept this contribution: it should first be done in the Google repository and then pushed back to the Git Hub repository.
We are in the process of doing 2 ways contributions, but this is not completed yet.

Issue #25, opened 2015-08-05
Review our Jetty xml files with respect to our context


ludoch:
Context class is at https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/master/appengine-jetty-managed-runtime/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java

Do we have duplicate settings in xml and this class?

gregw:
setAggregationBufferSize code can (or should) be done in XML rather than in code.   I suspect that the instancecheck is not working anyway.

gregw:
Fixed in PR #28

PR #28, opened 2015-08-06
Issue 25

gregw:
This is a partial fix for #25 
It creates a gae.mod module file that replaces all of the statically created ini files and the jetty.xml


gregw:
@ludoch This PR refines the way the DockerFile builds the image:
 * It no longer uses CURL to fetch jars from the maven repo, as that is slow on a slow network (come work from Australia :)
 * It uses the single jetty.version property set in the pom to trigger all the jetty dependencies and run scripts.  This has been updated to the latest 9.2.
 * It uses a new gae.mod and gae.xml to do the GAE specific setup, this allows jetty.xml etc. to be used unmodified (and thus easier to upgrade in future).
 * jetty_run.sh now sources a jetty_cmd.sh, which is dynamically generated in the DockerFile using the --dry-run, again making upgrades easier and avoiding hard coded dependencies.

gregw:
@ludoch Also, I note that the configuration was creating
````xml
    <New id="httpConfigForwarded" class="org.eclipse.jetty.server.HttpConfiguration">
````
but that this is not actually used anywhere?  I've copied it to gae.xml for now, but would like to delete it if it is not needed any more?

File: appengine-managed-runtime/pom.xml [hunk 1]
```diff
@@ -142,6 +142,32 @@
...
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>copy</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>com.google.appengine</groupId>
+                  <artifactId>appengine-api-1.0-sdk</artifactId>
+                  <version>${jetty.version}</version>
```
ludoch: I do not this it should be ${jetty.version} here.

buildimage.sh
```diff
@@ -2,11 +2,13 @@
 mvn clean install
 rm -rf docker/lib/ext
 mkdir -p docker/lib/ext
+cp appengine-jetty-managed-runtime/target/jetty-distribution.tar.gz docker/
```
ludoch: Would it be better to untar outside of the Docker build step? Since Docker is using copy on write, the image would otherwise contain the tar.gz and it exploded content...
The goal is to try to always minimize the image size.

gregw:
Updated appengine-api version to a property

Now expand the jetty distribution in the build script... although we were deleting it from the image anyway... but is clearer this way.  Eventually I would like to separate out a jetty home and jetty base, also for clarity.

ludoch:
Almost there. I do not see yet the expansion of jetty distribution in the build script.
I can approve this pr and we can do that in another one?

ludoch:
LGTM. Thanks!

PR #31, opened 2015-08-12
Issue 15

### LINKED ISSUE #15: StubSession shouldn't throw RuntimeException
I have an App Engine sandbox application that doesn't use sessions (<sessions-enabled>false</sessions-enabled>). It works perfectly.

On Managed-VM, I get the following exception: 
```java.lang.RuntimeException: Session support is not enabled in appengine-web.xml.  To enable sessions, put <sessions-enabled>true</sessions-enabled> in that file.  Without it, getSession() is allowed, but manipulation of sessionattributes is not.```

thrown by https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/master/appengine-jetty-managed-runtime/src/main/java/com/google/apphosting/vmruntime/jetty9/StubSessionManager.java#L100

To make my app work on MVM, I'm forced to activate sessions, thus creating more load and more writes on the datastore. 
Strangely, the problem doesn't happen when I access through the appspot url. The session set or remove calls are probably generated by frameworks (using Jersey + JSP + JSTL) but the call success aren't necessary to the app.

Thanks

janbartel:
This pull request depends on #25 and fixes #15.


ludoch:
LGTM

PR #36, opened 2015-08-14
Issue 32 and Issue 34

janbartel:
Changes to docker/etc/webdefault.xml to use correct JspServlet class, and to correct the security-constraints for disabling TRACE method. These changes need to be applied to all other webdefault.xml instances used by gae/gcloud.


ludoch:
LGTM

Issue #33, opened 2015-08-13
Remove Commit Delay


gregw:
remove the commit delay mechanism

gregw:
I've a bit stalled on other issues, so I'd like to work on this issue.   So I've assigned to self.

Issue #23, opened 2015-08-05
Servlet 3.1 Async feature does not work in Managed VM, all servlets in webdefault.xml have async:false


ludoch:
From a customer:

The main problem is in Google Filters. They do not support asyncRequest attribute. It prevents async feature of Servlet 3.0. Default web.xml should looks like:

    <filter>
        <filter-name>AbandonedTransactionDetector</filter-name>
        <filter-class>
            com.google.apphosting.utils.servlet.TransactionCleanupFilter
        </filter-class>
        <async-supported>true</async-supported>
    </filter>

    <filter-mapping>
        <filter-name>AbandonedTransactionDetector</filter-name>
        <url-pattern>/*</url-pattern>
    </filter-mapping>

    
    
    
    
    <filter>
        <filter-name>SaveSessionFilter</filter-name>
        <filter-class>
            com.google.apphosting.runtime.jetty9.SaveSessionFilter
        </filter-class>
        <async-supported>true</async-supported>
    </filter>

    <filter-mapping>
        <filter-name>SaveSessionFilter</filter-name>
        <url-pattern>/*</url-pattern>
    </filter-mapping>

    <filter>
        <filter-name>_ah_ParseBlobUploadFilter</filter-name>
        <filter-class>
            com.google.apphosting.utils.servlet.ParseBlobUploadFilter
        </filter-class>
        <async-supported>true</async-supported>
    </filter>

    <filter-mapping>
        <filter-name>_ah_ParseBlobUploadFilter</filter-name>
        <url-pattern>/*</url-pattern>
    </filter-mapping>

From Jeremy:
The filters in particular need to support async so that the servlets they wrap can as well. This changes the design pattern as the old
  try {
    // prepare
    chain.doFilter();
 } finally {
    // clean up
 }

pattern can't be used. Instead, the filter needs to detect if async has been started in the return path and/or whether this is an async dispatch in the prepare part. It may also need to add listeners to the AsyncContext in case the request gets completed in the background. 

It's all quite messy. It may be easier to convert some of the Filters to ServletRequestListeners.

The other issue is that the request may be processed on multiple threads so the ApiProxy context can't rely on a simple ThreadLocal. Specifically, AsyncContext#start "may" pass contextual information when executing the Runnable and we probably should so that applications can make API calls from background tasks. Doing this from user-land is messy so we should handle it within the container.

gregw:
[...]

PR #41, opened 2015-08-19
Async support

gregw:
This is not yet ready for an actual pull, but I wanted to create the PR so there is visibility of the direction these changes are taking.  Also this needs some unit tests!

This also includes the changes for removing the delayed commit wrappers.


gregw:
This branch contains progress on both #33 and #23, but currently need some assistance working out how to handle VmRuntimeUtils.flushLogsAndAddHeader ?

Currently this method is called as a request dispatch returns.  It flushes the API logs and adds a response header with the number of logs generated.  If the response is already committed, then it produces a warning that logs could have been lost.

With #33, response committing will no longer be able to be delayed. With #23, I have moved this call to the onComplete callback of AsyncContext.  Currently if the response is committed, it just flushes the logs.  If the log is not committed, then it proceeds as before.

So do we still need the header at all?  Is there anything else that needs to be done for logging when we get rid of commit delays?

isdal:
When we remove the commit delay we no longer need to set the log count header.

jboynes:
We're changing the way logging works to make things simpler. Rather than call the LogService, we are now writing logs directly from the MVM where they are picked up by a fluentd instance which handles sending them to Cloud Logging. All the LogService calls should be removed before disabling commit delay.

We're still defining exactly how the logs should be sent to fluentd. At the moment, best practice is to write the access/request log in W3C Common format and the diagnostic/error log in JSON format - support code is [here](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/master/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java) or I have a newer prototype [here](https://github.com/jboynes/appengine-runtime-java8/blob/master/appengine-java8-logging/src/main/java/com/google/appruntime/logging/JsonFormatter.java).

gregw:
@jboynes  by "All the LogService call", do you mean that not even a flush will be required when a request lifecycle is complete?

ludoch:
This is how I understand we should be doing for removing the old way of handling logs via the Logservice (V1 APIS):

1/ Remove these files:
- VMAppLogWriter.java
- VMRuntimeLogHandler.java
2/ trap all logservice APIs calls and store the entries vis the standard VMRuntimeFileHandler.java:
For that, changes needed in VmApiProxyEnvironment.java, 
3/ Remove the code referring to LOG_FLUSH_COUNTER_HEADER in VMRuntimeUtils.java

We will review the code with Tomas and Anthony.

apolito:
Followup:
The log flush code paths exist because appengine apis cannot be get 
VMRuntimeLogHandler logs to the logservice APIs  which we don't want to do in v2
VMRuntimeFileLogHandler logs to files that will get logged through the cloud service apis (what we want to use moving forward)

gregw:
This has been updated to use jetty-9.3.3, so the async unit test now passes.
This is now based off the jetty93 branch, so we should probably merge to there rather than master

gregw:
replaced by #67

PR #42, opened 2015-08-19
Adding Unit test utilities classes and some initial unit tests. WIP.

ludoch:


ludoch:
Greg: most of the remaining work is to correctly hook up the JettyRunner process from Junit. We cannot replicate what we did in our internal repo. I rely on you to decide the best way for the unit test base class to start a Jetty process on a port we calculate, which is as close as the real Jetty we are running in prod.

ludoch:
Added a web app test project and hook it to the tests of the Jetty module.
While working, some are failing, and it might take to long to run at each change.

ludoch:
Hi, 
Some tests you could look at are:
For JSP:  testJspNotCompiled in VmRuntimeJettyKitchenSinkTest 

For Session Management and testSsl_NoSSL In VmRuntimeJettyKitchenSink4Test
Thanks!

ludoch:
Merging now

janbartel:
Ludo,re the tests for sessions:  The testwebapp WEB-INF/appengine-web.xml has sessions enabled, so any tests that test for sessions being disabled should not work.  Given the way the tests are written, the appengine-web.xml is applied in the setup() step, which is the same for all tests in a test class. Thus, if the appengine-web.xml has sessions enabled, they will be enabled for every specific test in the class, so I don't see how testSessionsDisabled and testSessionsEnabled tests can live in the same class.

The other thing I noticed is that the VmRuntimeJettySessionTest class is setting this system property, I presume to change which appengine-web.xml file is used:
 System.setProperty("appengine_web_xml", "WEB-INF/sessions-appengine-web.xml");

However, I don't see  a sessions-appengine-web.xml file in a WEB-INF directory anywhere ... the only place it exists is in testwebapp/src/main/java/com/google/apphosting/tests/usercode/testservlets/sessions-appengine-web.xml. So presumably this is not being used. I verified that by changing WEB-INF/appengine-web.xml and those changes were picked up.  So, 2 questions from this:  1) should the sessions-appengine-web.xml file be moved into WEB-INF   and  2) it has sessions enabled anyway, so both appengine-web.xml and sessions-appengine-web.xml have sessions enabled - one of them should disable sessions, no?

Happy to make required changes, after getting your feedback.

cheers
Jan

janbartel:
Ludo, I went ahead and opened pull request #49

PR #44, opened 2015-08-21
Jetty93

gregw:
I've taken the jetty93 branch further to build up a docker image in stages (whose names and locations almost certainly needs to change):

 gce-debian-java8 - updates wheezy with java8 and certificate fix
 gce-jetty - updates gce-debian-java8 with a pure jetty-9.3 installation in the same style as the jetty-docker official image.
 appengine-jetty-managed-runtime - now builds its own image based on gce-jetty

All these docker images are built using the docker maven plugin attached to the install phase.  So you can build and install everything with a "mvn install" at the top level.   However a "mvn compile" does not install any images.

maven is mostly used for dependencies, except I cannot inject the image dependencies into Dockerfile

This is still a bit of a work in progress... but can be merged to jetty93 branch if you think it looks like the right direction.


ludoch:
LGTM

Issue #56, opened 2015-09-01
Pure Jetty9 image: standardize with the /app app location for Google Cloud


ludoch:
All the other GAE v2 runtimes seem to use the /app area for adding the user application.

We do it in the jetty9-compat image as well, but I think we should do it in the pure Jetty9 image by adding:
RUN ln -s /var/lib/jetty/webapps/root /app
WORKDIR /app

This way the user web app in /app is registered with the root context (i.e /) at the top of the Jetty server.
So far, we do not anticipate to support multiple web apps served by the same Jetty process. This might change in the future, but the first app and default one should be served at /

gregw:
It may be worthwhile to remove the deployer from the GAEv2 image.  Currently it is scanning $JETTY_BASE/webapps for multiple webapps to deploy.  This can take time, CPU and will probably break if somebody does drop another webapp in there (as it will not be a GAE anointed one).

So two proposals:  

Initially we should remove the deployer from GAE setup and replace with a gae-root-deploy module that does a hard coded (in xml) deployment of only the root webapp.

If eventually multiple webapps are desired, then we should create a gae-deploy module that creates an enhanced deployer that will deploy a normal war or webapp directory as a GAE context.  It there will be no need for a specific context.xml file to create the context as a GAE one.

Issue #55, opened 2015-09-01
pure Jetty9 image: tmp area is readonly


ludoch:
See an Admin console log after a deployment:
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.deploy.providers.WebAppProvider.fileAdded(WebAppProvider.java:438)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.deploy.providers.ScanningAppProvider.fileAdded(ScanningAppProvider.java:180)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.deploy.DeploymentManager.addApp(DeploymentManager.java:147)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.deploy.DeploymentManager.requestAppGoal(DeploymentManager.java:499)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.deploy.AppLifeCycle.runBindings(AppLifeCycle.java:188)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.deploy.bindings.StandardStarter.processBinding(StandardStarter.java:41)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.webapp.WebAppContext.doStart(WebAppContext.java:510)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.webapp.WebAppContext.preConfigure(WebAppContext.java:474)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.webapp.WebInfConfiguration.preConfigure(WebInfConfiguration.java:69)
A gae-default-jetty9ffg2tt-bph2 2015-09-01 08:23:17.000 at org.eclipse.jetty.webapp.WebInfConfiguration.resolveTempDirectory(WebInfConfiguration.java:306)
[...]

Issue #54, opened 2015-09-01
pure jetty9 image should support jsp/jstl out of the box


ludoch:
Kinda good default for Web Apps.

ludoch:
What other Jetty9 modules should be enabled out of the box?
JNDI? 
Or what is the guideline to easily enable module with a simple dockfile extension? Need documentation.

gregw:
I've recently added such documentation to the official docker image: see https://github.com/docker-library/docs/tree/master/jetty#configuration    Where is the best location for such doco on the pure jetty9 image? in a gce-jetty/README.md ?

I think JSP is a good default module, but I don't think JNDI is.

ludoch:
OK, and the gce-jetty/README can point to the official README.

PR #59, opened 2015-09-03
Fix for #54, #55, #56 issues.

ludoch:


gce-jetty/src/main/docker/Dockerfile
```diff
@@ -27,11 +27,14 @@ RUN mv /usr/local/jetty-distribution-* $JETTY_HOME && chown -R jetty:jetty $JETT
...
 RUN mkdir -p $JETTY_BASE
 WORKDIR $JETTY_BASE
-RUN java -jar $JETTY_HOME/start.jar --add-to-startd=http,deploy,setuid && \
+RUN java -jar $JETTY_HOME/start.jar --add-to-startd=http,deploy,jsp,jstl,setuid && \
     chown -R jetty:jetty $JETTY_BASE && \
-    mkdir /tmp/jetty
+    mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty
+
+RUN ln -s $JETTY_BASE/webapps/root /app
+WORKDIR /app
```
gregw: Why make the workdir /app?
This is a deviation from the official jetty-docker image that may hurt portability of webapps (albeit badly written ones that assume working dir).

ludoch: For the V1 compat mode, I think that is what we have specified. So yes, for
compat.
I guess, I can set the WORKDIR only in the jetty9-compat image.

gregw: I think for the pure jetty image, it is best to have the workdir as jetty_base, as derived images will have to customize that anyway. 

I'll accept and fix.

gregw: Not also that the official jetty-docker image now uses a small shell script as the entry point so that it is easy to run with a bash shell.  I think it is worthwhile to mimic that approach as it will give us better doco links.

For GAE it is not a problem as we introduce our own start script anyway!

Issue #62, opened 2015-09-04
GAEv2 deploy root only


gregw:
Rather than run the normal jetty deployer (which scans using time and CPU), a gae-root-deploy module should be created that does a hard coded (in xml) deploy of the root context.

PR #63, opened 2015-09-04
Jetty93 Fixes for #61 and #62

### LINKED ISSUE #61: Pure Jetty 9 image should mimic official docker-jetty entry point and documentation
The pure jetty 9 image should have the same entry point as the docker-jetty official image.
The similarities and customization should be documented in a README.md

gregw:
Updated Jetty93 Branch to fix #61 and #62 

The gce images now contain README.md files with some minimal documentation.

All the images now follow the docker best practise of having an entry point that can accept either other commands (eg bash) or arguments (eg --module=foobar) for the default entry point.

Removed the general deployer from the GAE image and replaced with a single webapp deployer.

### DIFF
--- a/appengine-jetty-managed-runtime/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppDeployer.java
+++ b/appengine-jetty-managed-runtime/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppDeployer.java
@@ -1,0 +1,96 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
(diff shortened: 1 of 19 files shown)


appengine-jetty-managed-runtime/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppDeployer.java
```diff
@@ -0,0 +1,96 @@
...
+import org.eclipse.jetty.server.handler.ContextHandlerCollection;
+import org.eclipse.jetty.util.annotation.Name;
+import org.eclipse.jetty.util.component.AbstractLifeCycle;
+import org.eclipse.jetty.util.resource.Resource;
+import org.eclipse.jetty.webapp.WebAppContext;
+import org.eclipse.jetty.xml.XmlConfiguration;
+
+public class VmRuntimeWebAppDeployer extends AbstractLifeCycle
+{
+  private final ContextHandlerCollection contexts;
```
ludoch: Will reformat later, but our conventions are { at the end of the code line, and {} even for 1 line if/else statements.

gregw: Trying very hard to comply with that.... have created and eclipse IDE that I think complies with your coding standards... but my subconscious obviously is still rebelling against unbalanced curlies and sneaks a few in still :)

PR #68, opened 2015-09-09
Put versions in docker tags

gregw:
This uses docker labels/tags for the version is a similar way to the official docker images


appengine-jetty-managed-runtime/pom.xml
```diff
@@ -217,13 +217,24 @@
...
           <execution>
-            <id>build</id>
+            <id>jetty-build</id>
             <phase>install</phase>
             <goals>
               <goal>build</goal>
             </goals>
             <configuration>
-              <imageName>appengine-mvn-opensource</imageName>
+              <imageName>gce-jetty:build</imageName>
```
gregw: Because the Dockerfile cannot be parameterized, we create a temporary image labelled build based on the jetty.version.  The Dockerfile then has gce-jetty:build hard coded in it.

appengine-jetty-managed-runtime/pom.xml
```diff
@@ -233,7 +244,18 @@
...
+          </execution>
+          <execution>
+            <id>gae-latest</id>
+            <phase>install</phase>
+            <goals>
+              <goal>build</goal>
+            </goals>
+            <configuration>
+              <imageName>appengine-mvn-opensource:latest</imageName>
+              <baseImage>appengine-mvn-opensource:${project.version}</baseImage>
```
gregw: This labels the appengine-mvn-opensource just built as latest, so that the examples work without needing to specify a particular version.

PR #69, opened 2015-09-10
Configure setup for jsp engine for tests

janbartel:
Use the jsp ServletContainerInitializer to configure the jsp engine for unit tests.


ludoch:
LGTM

PR #70, opened 2015-09-10
Implement changing session id

janbartel:
The Session.renewId(Request) method has been implemented for appengine compat sessions. This makes the Request.changeSessionId() method now work for appengine.

The only piece missing is the automatic changing of session id when a user authenticates - not sure that is appropriate as appengine does not use sessions as part of the authentication mechanism.


ludoch:
Nice! LGTM

PR #76, opened 2015-09-11
Set ENTRYPOINT to execute java.

jboynes:
This allows downstream images to just use CMD to run vanilla Java code.
Alternatively, they can override ENTRYPOINT to automatically run framework code.

### DIFF
--- a/gce-debian-openjdk/pom.xml
+++ b/gce-debian-openjdk/pom.xml
@@ -42,12 +42,12 @@
             <goals>
               <goal>build</goal>
             </goals>
-	    <configuration>
-	      <imageName>gce-debian-openjdk:8-jre</imageName>
-	      <dockerDirectory>${project.basedir}/src/main/docker</dockerDirectory>
-	    </configuration>
-	  </execution>
-	</executions>
+            <configuration>
+              <imageName>gce-debian-openjdk:8-jre</imageName>
+              <dockerDirectory>${project.basedir}/src/main/docker</dockerDirectory>
+            </configuration>
+          </execution>
+        </executions>
       </plugin>
     </plugins>
   </build>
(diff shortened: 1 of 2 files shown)


ludoch:
LGTM

gregw:
This is not really in line with the best practises for docker images.
I proposed a similar change to the official docker-jetty image (see https://github.com/appropriate/docker-jetty/pull/16) and the feedback is that best practise is to have an entrypoint that always allows a simple:
```
  docker run -it bash
```
This allows easy inspection of an image for testing/planning etc.

So having java as the entrypoint is going to be a surprise for many and against best practise.

The approach most often used is to have a small script as the entry point that determines if the args are a commands or arguments (in this case it would have to look for java class as well), and decides what to exec.

So perhaps it is simplest just to have null entry point and a CMD of bash?

jboynes:
@gregw The [best practice](https://docs.docker.com/articles/dockerfile_best-practices/#entrypoint) docs talk about:

> The best use for ENTRYPOINT is to set the image’s main command, allowing that image to be run as though it was that command (and then use CMD as the default flags).

The idea here was to enable:
```bash
docker run java com.example.Main
```
as outlined in that document.

More commonly, given you need to add code to run, you'd end up using a Dockerfile like:
```Dockerfile
FROM appengine/java
ADD app.jar .
CMD [ "-jar", "app.jar"]
```

You can always override the ENTRYPOINT on the command line to execute bash in order to inspect the image content. That would a less common usage than using it to run stuff though so I would not optimize for that.

ludoch:
It is also always possible to exec bash in a running container based on it's ID:

    docker exec -it  8fbff28f2b2b bash

jboynes:
From another perspective, the [stock Docker image](https://github.com/docker-library/java/blob/master/openjdk-8-jre/Dockerfile) for java does not set any ENTRYPOINT or CMD.

We could just remove this entirely.

Issue #79, opened 2015-09-15
README instructions are broken


waprin:
Following the README leads to an error on the docker build step.

Step 4 : ADD jetty /var/lib/jetty
INFO[0000] jetty: no such file or directory           

./buildimage.sh worked

gregw:
The README has been replaced by the merge of the jetty93 branch.
There is now no buildimage.sh script and all images are built via maven

PR #80, opened 2015-09-15
Update README To Just use buildimage script

waprin:
Fix for issue #79. One possible suggestion to keep README working is just to suggest the script.


ludoch:
LGTM

Issue #73, opened 2015-09-11
jetty93 VmRuntimeJettyAuthTest#testAuth_UntrustedInboundIp failing


gregw:
The VmRuntimeJettyAuthTest#testAuth_UntrustedInboundIp test in jetty93 branch is failing. It appears to be trying to contact the real authentication service and hanging.

gregw:
The test has been commented out for now to get a clean build

ludoch:
I think this is because 
    getSecurityHandler().setAuthenticator(new VmAppEngineAuthenticator(
        this, getSecurityHandler().getAuthenticator()));
was removed in VmRuntimeWebAppContext.java

ludoch:
ignore preview comment. still looking

PR #86, opened 2015-09-18
Fixed #78 and #73

### LINKED ISSUE #78: Tests Fail On Clean Checkout
VmRuntimeJettyAuthTest.testAuth_UntrustedInboundIp:154 expected:<307> but was:<403>
  VmRuntimeJettyAuthTest.testAuth_UntrustedInboundIpWithQuery:165 expected:<307> but was:<403>

gregw:
Cleaned up poms to clearly use java 1.8
Use 127.0.0.2 as force remote IP, which is known to be untrusted.

### DIFF
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRequestUtils.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRequestUtils.java
@@ -38,6 +38,9 @@ public class VmRequestUtils {
       return true;
     } else if (remoteAddr.startsWith(LINK_LOCAL_IP_NETWORK)) {
       return true;
+    } else if (remoteAddr.equals("127.0.0.2")) {
+      // Test untrusted.
+      return false;
     } else if (remoteAddr.startsWith("127.0.0.")) {
       // Allow localhost.
       return true;
--- a/appengine-jetty-managed-runtime/pom.xml
+++ b/appengine-jetty-managed-runtime/pom.xml
@@ -27,8 +27,6 @@
   <url>http://maven.apache.org</url>
   
   <properties>
-    <maven.build.timestamp.format>yyyyMMddHHmm</maven.build.timestamp.format>
-    <jetty.version>9.3.2.v20150730</jetty.version>
     <appengine.api.version>1.9.25</appengine.api.version>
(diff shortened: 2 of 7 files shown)


gregw:
fix for #78 and #73

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRequestUtils.java
```diff
@@ -38,6 +38,9 @@ public static boolean isTrustedRemoteAddr(boolean isDevMode, String remoteAddr)
       return true;
     } else if (remoteAddr.startsWith(LINK_LOCAL_IP_NETWORK)) {
       return true;
+    } else if (remoteAddr.equals("127.0.0.2")) {
```
jboynes: Curious about the semantics required here. [RFC5735](https://tools.ietf.org/html/rfc5735) says:
>127.0.0.0/8 - This block is assigned for use as the Internet host loopback address.  A datagram sent by a higher-level protocol to an address anywhere within this block loops back inside the host.  This    is ordinarily implemented using only 127.0.0.1/32 for loopback.  As   described in [RFC1122], Section 3.2.1.3, addresses within the entire   127.0.0.0/8 block do not legitimately appear on any network anywhere.

Which means we should either trust all "127." addresses in the block, or only the "ordinary implementation" of "127.0.0.1" but treating just 127.0.0.2 as special seems odd.

ludoch:
LGTM

jboynes:
@ludoch thoughts on trusted ips ?

ludoch:
Please, add your concerns.

jboynes:
@ludo My [prior comment](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/pull/86#discussion_r39869695) above about making 127.0.0.2 special

gregw:
@jboynes The previous impl was using the hosts local address and assuming that would not be trusted.  Yet it had a devenv setup that trusted all localhosts.

Thus I just added a single exclusion to mark 127.0.0.2 as untrusted and left the 127.0.0. as trusted.
I could not see any other 127 addresses being used, so it should be find to change this to just trust 127.0.0.1, but that is essentially equivalent to what we have now.

Issue #91, opened 2015-09-24
Update Logging to new logging architecture


gregw:
Remove the logging that was tied to the delayed commit mechanism and replace with the JSON formatted java logger.

gregw:
@jboynes @ludoch 
In the current VmRuntimeWebAppContext there is the following code 
```
      if (!VmRuntimeUtils.waitForAsyncApiCalls(requestSpecificEnvironment,response)) {
        logger.warning("Timed out or interrupted while waiting for async API calls to complete.");
      }
```
Just before it does the log flushing.  This was in the delayed commit handling, but I have moved it to the onComplete async handling in the async-support branch.

My question is - do we still need to wait for async API calls to complete?  Was this done only so that all the logs could be flushed with the response headers?  Now that we don't delay commit and will support async request handling, is there any reason to delay the end of the request lifecycle until all AsyncAPICalls have completed?     Note that if they do try to use the servlet async API after onComplete, they will get ISE thrown (as they should).

jboynes:
For now I think this is still needed as there is a window between when the response is written and when the request is completed in which API calls could still be made. With the current serving path API calls made in that window may error; this is why we wait for all logging calls to complete before committing the response. There are changes to the serving path that will remove that limitation but those are not yet released.

gregw:
@jboynes but I'm asking this in the context of the async-supported branch where we have already removed the delayed commit.  Thus the response is probably already committed and it is not possible to add headers - as the waitForAsyncApiCalls attempts to do?

gregw:
I've tried rerouted the jetty logging mechanism to java.util.logging only to decide that the jetty JavaUtilLog class is rather deficient (wrong source class/method name for example).     If reimplemented that class and it will be release in 9.3.4 in about a weeks time.

gregw:
This is done in the async-support branch waiting for merge

gregw:
fixed in async-support branch

PR #92, opened 2015-09-24
Add a separate module for AppEngine logging

jboynes:
Add a separate logging module - main runtime will be updated to use it in a separate PR.


gregw:
LGTM and needed for  #91

PR #99, opened 2015-09-25
Build cleanup.

jboynes:
Name all the modules and define dependencies between them.
Build each docker iamge in the package phase and add tags in each module.


ludoch:
Thanks! LGTM

Issue #98, opened 2015-09-25
Automate adding all jars from dependencies


jboynes:

gregw:
I'll take this one if you like.  You can do #97

jboynes:
OK

gregw:
with the explicit fix from #95 the dependencies copied were:
```
appengine-api-1.0-sdk-1.9.25.jar            
appengine-jetty-managed-runtime-1.9.26-SNAPSHOT.jar  
gson-2.3.1.jar
appengine-java-logging-1.9.26-SNAPSHOT.jar  
appengine-managed-runtime-1.9.26-SNAPSHOT.jar
```

Using the copy-dependencies target and excluding tests, jetty and it's transitive dependencies gives:
```
appengine-api-1.0-sdk-1.9.25.jar
appengine-java-logging-1.9.26-SNAPSHOT.jar
appengine-managed-runtime-1.9.26-SNAPSHOT.jar
commons-codec-1.10.jar
commons-logging-1.2.jar
gson-2.3.1.jar
httpclient-4.4.1.jar
httpcore-4.4.1.jar
```

I think the additional dependencies are valid as they are dependencies of the runtime.  Here is the output of dependency:tree:
```
[INFO] ------------------------------------------------------------------------
[INFO] Building appengine-jetty-managed-runtime 1.9.26-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ appengine-jetty-managed-runtime ---
[INFO] com.google.appengine:appengine-jetty-managed-runtime:jar:1.9.26-SNAPSHOT
[INFO] +- com.google.appengine:appengine-api-1.0-sdk:jar:1.9.25:compile
[INFO] +- com.google.appengine:appengine-managed-runtime:jar:1.9.26-SNAPSHOT:compile
[INFO] |  +- com.google.code.gson:gson:jar:2.3.1:compile
[INFO] |  +- org.apache.httpcomponents:httpcore:jar:4.4.1:compile
[INFO] |  +- org.apache.httpcomponents:httpclient:jar:4.4.1:compile
[INFO] |  +- commons-logging:commons-logging:jar:1.2:compile
[INFO] |  \- commons-codec:commons-codec:jar:1.10:compile
[INFO] +- org.eclipse.jetty:jetty-security:jar:9.3.2.v20150730:compile
[...]

PR #100, opened 2015-09-25
Automate dependencies #98

gregw:
Fix for #98.   

This uses a moderately ugly excludeGroupIds list that could become and issue if dependencies get complex.

### DIFF
--- a/appengine-java-logging/pom.xml
+++ b/appengine-java-logging/pom.xml
@@ -1,16 +1,19 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-  ~ Copyright 2015 Google Inc. All Rights Reserved.
-  ~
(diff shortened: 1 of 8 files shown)


ludoch:
we already have a repackaged gson in the non jetty dependent jar (using jarjar, see its pom)
I think the better solution is to generalized our deps via jarjar (and possible do a single jar of deps with jarjar) to not interfere with user code.

gregw:
I'm not a big fan of repackaging tools like jarjar.   They are OK for very specific usages, but for creating artefacts that are intended for further extension and/or composition the combined jar approach tends to hide dependencies and often can result in duplicate classes.

I'm going to close this PR and reopen from a branch (rather than master), so we can discuss more.

gregw:
Moved to a new PR #104

PR #101, opened 2015-09-26
Make clean target remove docker images

jboynes:
I only managed to get this working by removing the explicit tag names. This kind of made sense to me for use during the development process but probably is not what we want for release builds. I'll open an issue to discuss.

### DIFF
--- a/appengine-jetty-managed-runtime/pom.xml
+++ b/appengine-jetty-managed-runtime/pom.xml
@@ -232,11 +232,7 @@
               <goal>build</goal>
             </goals>
             <configuration>
-              <forceTags>true</forceTags>
-              <imageName>appengine-mvn-opensource:${project.version}</imageName>
-              <imageTags>
-                <imageTag>latest</imageTag>
-              </imageTags>
+              <imageName>appengine-mvn-opensource</imageName>
               <dockerDirectory>${project.basedir}/src/main/docker</dockerDirectory>
               <resources>
                 <resource>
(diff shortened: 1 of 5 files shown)


appengine-jetty-managed-runtime/pom.xml
```diff
@@ -247,6 +243,18 @@
...
           </execution>
+          <execution>
+            <phase>clean</phase>
+            <goals>
+              <goal>removeImage</goal>
+            </goals>
+            <configuration>
+              <imageName>appengine-mvn-opensource</imageName>
+              <imageTags>
+                <imageTag>latest</imageTag>
```
gregw: Removing latest is not guaranteed to remove the artefacts created by this build/branch.
In development, if multiple branches are being developed it may well be likely that java 7 and java 8 images are both deployed,   Deleting the latest only deletes the last built and is not related to this poms version in any way.

jboynes: This change no longer explicitly tags changes and as a result they automatically get tagged as "latest" but the plugin does not allow that for the removeImage goal - to remove "latest" the <imageTags> configuration needs to be specified. 

IOW, it does remove ones that were just built. If you have other tags for the image those will not be removed.

gregw:
I'm closing this one for now, and creating a new PR for #102 that is cleaner in how it handles docker version tags and also does the removeImage during clean

PR #104, opened 2015-09-26
Use maven copy dependency plugin for #98

gregw:
Use maven copy dependency plugin for #98

This is the same as #100 except rebased on latest master and from a branch so that discussion can take place during other development.

### DIFF
--- a/appengine-java-logging/pom.xml
+++ b/appengine-java-logging/pom.xml
@@ -13,13 +13,11 @@
   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
   See the License for the specific language governing permissions and
   limitations under the License.
-
-  -->
+-->
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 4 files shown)


ludoch:
There are some good reasons why we are using jarjar (at least for our GAE V1 runtime) where some customers had conflicting versions in their app versus the runtime.
For Managed VMs, I guess this is fine as we are using the default Jetty System, Server and App Classloaders?
Just want to make sure we are not creating possible regressions for existing apps in compat  mode.

gregw:
Perhaps in V1 the classloader was not able to hide the conflicting jars.
With the default Jetty classloader we have the ability to hide classes and packages.  However, by default only org.eclipse.jetty packages are hidden as server classes.

So regardless of how we package these dependencies, we need to decide which are hidden and which are exposed to the webapplication.      Ideally this list can also be auto generated.

gregw:
For now I have added to root.xml a fixed list of packages to be hidden from the webapp to avoid clashes with any classes supplied by the webapp:
```xml
  <!-- Hide the additional server implementation libraries -->
  <Call name="addServerClass"><Arg>org.apache.commons.codec.</Arg></Call>
  <Call name="addServerClass"><Arg>org.apache.commons.logging.</Arg></Call>
  <Call name="addServerClass"><Arg>org.apache.http.</Arg></Call>
  <Call name="addServerClass"><Arg>com.google.gson.</Arg></Call>
```

gregw:
merged with PR #108 to pass circleci

Issue #96, opened 2015-09-25
Cloud debugger does not seem to work with the jetty9 compat image


ludoch:
I check the internal docker image we have, and Cloud debugger is working there.
But not from what we have in github.
The script diff I see is only this:
# Test if the argument is executable and if then run it directly
if type "$1" &>/dev/null; then
  exec "$@"
fi

in the jetty_run.sh script
Not sure what else is different..

gregw:
Those extra lines are to conform to the docker policy of supporting 'docker run image bash' or similar.  If they are a problem for the debugger, then just remove them as they are not that important.

ludoch:
I tried briefly, and the container did not start (no response)
Did not have time to debug

gregw:
I'll be available this evening to help debug if need be.

ludoch:
1 issue at least is that wget is not in the base image. I'll add it to the base jdk image.

ludoch:
via ssh to the image, 
Info: Limiting Java heap size to: 2569M
/var/lib/jetty/jetty_run.sh: line 30: wget: command not found
/var/lib/jetty/jetty_run.sh: line 36: wget: command not found
Downloading Cloud Debugger agent from https://storage.googleapis.com/vm-config.openblend-demo.appspot.com/
/var/lib/jetty/jetty_run.sh: line 41: wget: command not found
tar (child): cdbg_java_agent.tar.gz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
/var/lib/jetty/jetty_run.sh: line 45: cdbg/format-env-appengine-vm.sh: No such file or directory
2015-09-25 17:06:31.606:INFO::main: Logging initialized @235ms
2015-09-25 17:06:32.239:INFO:oejs.Server:main: jetty-9.3.2.v20150730
2015-09-25 17:06:32.287:INFO:oejs.AbstractNCSARequestLog:main: Opened /var/log/app_engine/request.2015_09_25.log
2015-09-25 17:06:32.532:INFO:oejs.ServerConnector:main: Started ServerConnector@310ba866{HTTP/1.1,[http/1.1]}{0.0.0.0:8080}
Exception in thread "main" java.lang.reflect.InvocationTargetException
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
[...]

PR #105, opened 2015-09-27
Download gcloud debugger from known public location #96

gregw:
Following advice for work around for cloud debugger from #96

### DIFF
--- a/appengine-jetty-managed-runtime/src/main/docker/jetty_run.sh
+++ b/appengine-jetty-managed-runtime/src/main/docker/jetty_run.sh
@@ -18,7 +18,6 @@ HEAP_SIZE=$(awk -v frac=$HEAP_SIZE_FRAC -v res=$RAM_RESERVED_MB /MemTotal/'{
 echo "Info: Limiting Java heap size to: $HEAP_SIZE"
 
 DBG_AGENT=
-
 if [[ "$GAE_PARTITION" = "dev" ]]; then
   if [[ -n "$DBG_ENABLE" ]]; then
     echo "Running locally and DBG_ENABLE is set, enabling standard Java debugger agent"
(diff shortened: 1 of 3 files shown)


vlad-lifliand:
This is going to work, but not too well:
1. It is possible that different instances of the same version will be using different versions of the debugger agent. It will happen, for instance, if the number of instances is increased after a newer version of the debugger agent was released. We would like to avoid this kind of asymmetry.
2. Every instance just downloads file from the same URL, on every boot. This is just wasteful.
3. It is also more complicated than necessary. The solution we had for java7 was snapshotting the debugger agent version on deployment, but it doesn't work for custom runtimes.

I suggest simplifying the whole thing as following:
1. Add debugger agent binary to the Docker image by adding this line to Dockerfile:
ADD https://storage.googleapis.com/cloud-debugger/appengine-java/current/cdbg_java_agent.tar.gz /home/vmagent/cdbg
(it will also unpack files automatically)
2. jetty_run.sh just need the last part of formatting the command line:
DBG_AGENT="$( /home/vmagent/cdbg/format-env-appengine-vm.sh )"

ludoch:
Thanks Vlad!
Greg: let's do as Vlad said.

gregw:
@ludoch Should I install the cdbg_java_agent.tar.gz in the gce-openjdk, gce-jetty or appengine-mvn-opensource image?     I think gce-openjdk can have it installed?

gregw:
@vlad-lifliand  is /home/vmagent/cdbg a well known location?
There has been some discussion of adding other jars to the openjdk image and /usr/local/share/java or /usr/share/java may be a better place to install such libraries?

gregw:
@ludoch I'm guessing I should do the same for the profiler?
```shell
    CPROF_AGENT_URL="https://storage.googleapis.com/cloud-profiler/appengine-java/current/cloud_profiler_java_agent.tar.gz"
```

gregw:
I've updated the PR #105 so that it installs both the cloud debugger and cloud profiler into the gce-debian-openjdk image.    Both the jetty and gae images have been updated to use these installed versions.

Note that I don't yet know how to test that these work.

Also note circleci is failing until #108 is merged.

vlad-lifliand:
Good. I don't have any opinion where the debugger binaries should go. Note
that it's not just a jar file...

Issue #102, opened 2015-09-26
Tagging images during release builds


jboynes:
The build currently tags images in different ways:

* gce-debian-openjdk is tagged with "8-jre"
* gce-jetty is tagged with the Jetty version being used
* appengine-mvn-opensource is tagger with "latest" and the build version "1.9.26-SNAPSHOT"

In #101 I've changed that to always use "latest" as that is a mutable reference (i.e. it can and is expected to point to different images) which seems to make sense for development versions. IOW, when I develop I always want the "latest" stuff I just built.

However, this is not appropriate for a release build as it leads to ambiguity. The "latest" has special meaning to Docker in that it is the default tag pulled if no matching image is found in the local registry. If a matching image _is_ found, then no pull will be performed even if there is a different "latest" image in the origin registry; this is different to Maven's handling of "SNAPSHOT" versions.

When the image is built, the baseImage is set to the UUID of the actual image resolved by the FROM instruction. This fixes the chain of image dependencies so it will not change if images are later retagged.

AIUI, this makes tag names such as "8-jre" and "jetty-9.3.2" problematic because repeated builds of this project will result in different image chains if the underlying debian image changes. To avoid that, we could use the Maven version throughout e.g. gce-jetty:1.9.26-SNAPSHOT for a development build or gce-jetty:1.9.26 for a release build. However, I think that's confusing as there no cue on which OpenJDK or Jetty version was used.

A few options come to mind:

1. include the framework in the image name and use the project version as the tag i.e. gce-jetty9.3:1.9.26
1. separate the modules with their own version numbers e.g. gce-jetty:9.3.2.vYYYYMMDD where YYYYMMDD is the date _we_ built the image not the one of the Jetty release
1. combine e.g. gce-jetty-9.3.2:vYYYYMMDD with the date as above which may be a bit less confusing

All these options require specifying the parent tag in the Dockerfile
```
FROM gce-debian-openjdk:????
```
where `????` is an immutable reference. As we can't use "latest" here, this means pre-processing the Dockerfile to perform substitution.

Option 3. `gce-jetty-9.3.2:vYYYYMMDD` is the option I dislike least - suggestion for a better one would be welcome/

gregw:
I agree the tagging approach was inconsistent and slightly inflexible.

However, it was motivated by again trying to match the official docker releases which start from a base of java images with tags like: 7-jre, 8-jre etc.

[...]

PR #109, opened 2015-10-01
Clean docker images #102

gregw:
Added docker removeImage clean targets for 
Also templated the handling of the jetty version in the Dockerfile as a slight cleanup for #102


ludoch:
need to merge the conflict.

gregw:
The failure that CircleCI is having here appears to be a docker problem: https://github.com/docker/docker/issues/9939

Issue #82, opened 2015-09-16
Enable ALPN in openjdk8 image


gregw:
Enable ALPN in openjdk9 image as part of the boot classpath of the JVM.   Use this ALPN for HTTP/2 in the pure jetty image and GAEv2 images

lesv:
That's a great idea -- if you do it, remember to add in webapps/root.xml  (or it's equivalent)

```  
<Call name="addSystemClass">
    <Arg type="String">org.eclipse.jetty.alpn.</Arg>
</Call>
  <Set name="parentLoaderPriority">
    <SystemProperty name="jetty_parent_classloader" default="false"/>
</Set>
```
So ALPN can be found by Netty in gRPC.

gregw:
@lesv 
I think it should probably be exposed by default in jetty, so I've created this issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=477641

lesv:
Probably should, but it's definitely not now.  Thanks -- good idea.

gregw:
There are two ways we can proceed with this.

We are currently working with the openjdk9 developers reviewing their ALPN implementation.   The key thing about the proposed openjdk9 ALPN support is that it is a significantly different model to what we have implemented in the jetty project.  Specifically it has a different callback model that makes the selection of the cipher/protocol/alias an iterative process allowing co-dependencies.

We could just add the existing jetty ALPN to the openjdk8 images and leave the porting to the new model to when things are switched to openjdk9.  

Alternatively, we could do the work now to morph the current jetty image to offer the same API and interaction style as proposed for openjdk9.  This would both test the openjdk9 API (as it really needs real test cases to check) and would mean that newly developed ALPN dependent code would not need to be reworked in a year or two, which may make an impediment for openjdk9 adoption.

As ALPN is "new" in the context of the official google java images, I think consideration should be given to this second approach.

gregw:
@lesv any comment on the openjdk9 API style?

lesv:
I can talk about alpn-jetty interaction in how I use it, but @ejona86 is better equipped to comment.

ejona86:
[...]

PR #111, opened 2015-10-13
Added ALPN support to gce-debian-openjdk #82

gregw:
Issue #82
Moved gdbg, gprof and alpn to /opt
Added /opt/alpn/format-env-appengine-vm.sh create ALPN java args
Added $JETTY_BASE/modules/gce-alpn.mod to avoid downloading jetty ALPN

The env variable ALPN_ENABLE and ALPN_DISABLE follow the same pattern as
used for cdbg and cprof.

If ALPN_ENABLE is set, then the jetty_run.sh and docker-entrypoint.bash scripts
will call /opt/alpn/format-env-appengine-vm.sh.   That script will look for
ALPN_DISABLE and if not set, will produce arguments for the boot classpath.


ludoch:
Just need to resolve the conflict and good to go.

Issue #106, opened 2015-09-28
Circleci failing


ludoch:
Somebody played with circleci tooling on the project.
The first thing this CI system is calling

mvn dependency:resolve

This is leading to this build error:
...
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Google App Engine Runtime Parent ................... SUCCESS [  5.924 s]
[INFO] Google App Engine Logging Support .................. SUCCESS [  0.714 s]
[INFO] Google App Engine Image for OpenJDK ................ SUCCESS [  0.080 s]
[INFO] Google App Engine Image Jetty ...................... SUCCESS [  0.007 s]
[INFO] Google App Engine Runtime Support .................. FAILURE [  2.704 s]
[INFO] Compatibility test application ..................... SKIPPED
[INFO] Google App Engine Image for GAEv1 Java Compatibility SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 11.561 s
[INFO] Finished at: 2015-09-28T10:19:34+00:00
[INFO] Final Memory: 20M/1963M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project appengine-managed-runtime: Could not resolve dependencies for project com.google.appengine:appengine-managed-runtime:jar:1.9.26-SNAPSHOT: Could not find artifact com.google.appengine:appengine-java-logging:jar:1.9.26-SNAPSHOT in sonatype-google-releases (https://oss.sonatype.org/service/local/repositories/google-releases/content) -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR] mvn <goals> -rf :appengine-managed-runtime

gregw:
The problem is with [Multi-module Build Dependencies]( http://www.alexecollins.com/migrating-to-circleci/), will add a circle.yml to fix.

gregw:
[...]

PR #112, opened 2015-10-13
Untar with --no-same-owner for circleci #106

gregw:
Following advice from circleci


gregw:
[Bio-readouts are all in the green! looks like she's alive!](http://movie-sounds.org/quotes/aliens/Bioreadouts-are-all-in-the-green-Looks-like-shes-alive.mp3)

Issue #117, opened 2015-10-22
How to setup Cloud debugger and java opts passed via env with the OpenJDK8 image?


ludoch:
Soon, the Cloud SDK will generate this dockerfile when you wan to deploy a .jar file (for example, SpringBoot app listening to port 8080):

    FROM gcr.io/google_appengine/openjdk8
    ADD foo.jar /app/
    ENTRYPOINT ["java", "-jar", "/app/foo.jar"]

What is the best way to customize it with ENV variables passed via docker build/run and enabling jvm opts and cloud debugger/profile?

gregw:
@ludoch 

It would be better if it could be:
```
FROM gcr.io/google_appengine/openjdk8
ADD foo.jar /app/
ENTRYPOINT ["/docker-entrypoint.bash"]
CMD ["java","-jar","/app/foo.jar"]
```
As this would allow a docker-entrypoint.bash script similar to the one in gke-jetty to run and add the cdbg, alpn and cprof properties as needed.   It would also allow for docker commands like:
```shell
docker run theImage -verbose
```
to simply add JVM arguments
or commands like
```shell
docker run theImage bash
```
to explore the image.

I will write this entry-point now to better illustrate

PR #118, opened 2015-10-22
Cloud Debugger/Profiler Entry point #117

gregw:
This entry point can run java commands and derived images need only set the CMD.
If this is acceptable, we can look are removing the near duplicate in gke-jetty


gregw:
for #117

ludoch:
Lost track of the changes, but initialy, IIRC, we supported JAVA_OPTS env variable passed to the container (you can define env variables in the app.yaml that needs to be there to configure the app, and they are passed to the container). I think we should also keep it.

gregw:
@ludocj OK, will add JAVA_OPTS now

gregw:
@ludoch JAVA_OPTS support added

ludoch:
LGTM

PR #139, opened 2015-12-18
improve README for gke images

gregw:
Improve the documentation in the README for the GKE images.
This also contains comments and simplifications in the entry point slips.. which detected a bug in that the gke_env.sh script had not been fully updated for the /opt home for cdbg and cprof


ludoch:
Very nice, thanks!

Issue #141, opened 2015-12-21
Please rename folders to prevent confusion


arouzrokh:
gke-debian-openjdk -> openjdk8
gke-jetty-> jetty9

We will keep the version numbers as they are useful. 

README in each image directory should have full description of what versions of JDK or Jetty are included (major and minor version) even though this information can probably be discerned from the dockerfile. 

Thanks.

gregw:
@ludoch I think we need a more thorough review of the naming conventions, specially with the work that I'm currently doing for #140 which adds additional modules and images.

Currently in my branch for #140 I have the modules below and propose the renaming as follows:

Module | Artefact | Comment | New Name?
----------|----------|-------------|-------
gke-debian-openjdk | image | debian/java image suitable for GKE | gke-debian-openjdk8
gke-jetty | image | jetty image suitable for GKE | gke-jetty9
appengine-java-logging | jar | JsonFormatter and LogContext for java.util.logging.   Should be merged into appengine-managed-runtime and thus avoid the "java" in the name? | -
appengine-managed-runtime | jar | Non container specific runtime classes. | managed-runtime
jetty-managed-runtime | jar + zip | Container specific runtime classes. Produces a zip of a $JETTY_BASE directory capable of running compat webapps.  | managed-runtime-jetty9??
gae-jetty-managed-runtime | image | Uses managed-runtime-jetty9 and gke-jetty9 to create compat docker image | gae-managed-runtime-jetty9

This uses the gke- prefix to identify modules that produce images that can be deployed on GKE and gae- for the GAE images.   Other modules just have functional names scoped by the package com.google.apphosting

I'll proceed on this basis in my #140 work, but happy to rename again if there is a better scheme.

gregw:
Another complication is the docker image names themselves.   Currently gke-debian-openjdk produces a docker image of the same name with the label 8-jre, which matches the conventions for the standard docker images.

If we rename this module to gke-debian-openjdk8, then what should the label be?  Just "latest" or perhaps the specific version of openjdk8  eg u51 ?

ludoch:
See a start  of a PR at
https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/pull/146

ludoch:
gcr,io names are there:
https://cloud.google.com/appengine/docs/managed-vms/custom-runtimes

ludoch:
that would give these local image names

 docker images

[...]

Issue #140, opened 2015-12-20
Async support: create a distro in a tar gz file


ludoch:
We want to support both run and deploy (i.e create good staging content with correct quickstart web.xml) for the latest Jetty 9.3.x async branch.
We will need to create a distro we can make available as a downloadable file in this github project, that also will be (for now) bundled in the next Cloud SDK when possible.

- We want to keep the 2 directories jetty home and jetty base as 2 top level separated directories.
- we want a single main entry that will just generate a correct quickstart web.xml in the web-inf dir of a given web app directory archive.
Today, for this we created the quickstartgenerator.jar that has this main (passing a web directory) and has a deps on all correct *.jar are relative path in the jar manifest path. If there is a better way to either do this jar, or a way with java jar start.jar with correct flags to only generate the quick xml, let's do that and we will modify the staging phase that calls it (See current call at: https://code.google.com/p/googleappengine/source/browse/trunk/java/src/main/com/google/appengine/tools/admin/Application.java lines 717 calling line 1342). The long term plan is to move this code in a new SDK open source project. But for now (january 2016) I'll do the necessary changes in our google code, and bundle the correct distro in a new Cloud SDK deliverable.

We sill want to bundle the old jetty bundle in the Cloud SDK for a short transition period and it will be deleted later in Q1.
(current distro is in 
~/google-cloud-sdk/platform/google_appengine/google/appengine/tools/java/lib/jetty-base-sdk
and
~/google-cloud-sdk/platform/google_appengine/google/appengine/tools/java/lib/java-managed-vm
with the main generator in ~/google-cloud-sdk/platform/google_appengine/google/appengine/tools/java/lib/java-managed-vm/appengine-java-vmruntime/quickstartgenerator.jar

ludoch:
As soon as we have this in place, we will merge the async branch to master and that will be our default target.

gregw:
eta wednesday

gregw:
@ludoch  Sorry for the delay on this.   It is taking me a bit longer that I thought.

There are two issues making this difficult:
 * The jetty-base in the images is generated by running start.jar --add-to-startd from the Dockerfile
 * The jetty-base is built up as an overlay from the one in the image in the gke-jetty image

[...]

PR #147, opened 2016-01-05
Refactor modules for #140 to have a JETTY_BASE in a zip file

gregw:
This is a refactor of the modules for #140 so that there is a maven artefact that contains the entire jetty_base for the compat image.

This is not yet a complete PR as I need some feedback on how to make the jetty base directly usable.  Specifically how/where an app will be added to be deployed.

Some renaming has also been done to move this towards #141, but that is not complete.


gregw:
Closing in favour of newer version based on async with fixes

Issue #136, opened 2015-12-16
logging doesn't work locally


lesv:
In my appengine-web.xml I enable the following lines:

    <system-properties>
      <property name="java.util.logging.config.file" value="WEB-INF/logging.properties"/>
      <property name="log4j.debug" value="true" />
    </system-properties>

I've tried "/WEB-INF/logging.properties", and a full path "/Users/.../target/appengine-staging/WEB-INF/logging.properties".  The /WEB-INF says file not found, but the others work.

On my machine, in [my project](https://github.com/GoogleCloudPlatform/cloud-bigtable-examples/tree/master/java/managed-vm-gae), doing `mvn gcloud:run` I see:

Temporary staging for module default directory left in /Users/lesv/github/cloud-bigtable-examples/java/managed-vm-gae/target/appengine-staging
[INFO] Running python -S /Users/lesv/google-cloud-sdk/platform/google_appengine/dev_appserver.py --skip_sdk_update_check=true -A app /Users/lesv/github/cloud-bigtable-examples/java/managed-vm-gae/target/bigtable-hello-1.0-SNAPSHOT/app.yaml --jvm_flag=-Xbootclasspath/p:/Users/lesv/.m2/repository/org/mortbay/jetty/alpn/alpn-boot/8.1.3.v20150130/alpn-boot-8.1.3.v20150130.jar --runtime=java
[INFO] INFO     2015-12-16 23:50:27,763 application_configuration.py:431] No version specified. Generated version id: 20151216t235027
[INFO] INFO     2015-12-16 23:50:27,764 devappserver2.py:769] Skipping SDK update check.
[INFO] INFO     2015-12-16 23:50:27,818 api_server.py:205] Starting API server at: http://localhost:60036
[INFO] INFO     2015-12-16 23:50:27,827 dispatcher.py:197] Starting module "default" running at: http://localhost:8080
[INFO] INFO     2015-12-16 23:50:27,834 admin_server.py:116] Starting admin server at: http://localhost:8000
**[INFO] 2015-12-16 23:50:28.170:INFO::main: Logging initialized @326ms**
**[INFO] 2015-12-16 23:50:28.362:INFO:oejs.Server:main: jetty-9.2.10.v20150310**
**[INFO] 2015-12-16 23:50:28.376:WARN:oejsh.RequestLogHandler:main: !RequestLog**
[...]

Issue #124, opened 2015-11-10
java-compat from MVM's - what should I set to read logging.properties ??


lesv:
I've tried:
      `<property name="java.util.logging.config.file" value="logging.properties"/>`
and
      `<property name="java.util.logging.config.file" value="WEB_INF/logging.properties"/>`
and
      `<property name="java.util.logging.config.file" value="/app/WEB_INF/logging.properties"/>`

erugeri:
In my appengine-web.xml I have:

```
<system-properties>
	<property name="java.util.logging.config.file" value="WEB-INF/logging.properties"/>
</system-properties>
```

To check if it works, set the content of logging.properties like that:
```
.level=ALL
```

lesv:
I'm seeing file not found (in the log) for all three variations.

gregw:
@lesv 
I've found, fixed and forgotten this bug once already.   I'll refresh my brain and refix shortly.

gregw:
Depending on your environment, PR #148 may have only partially fixed this issue.    The fix in that PR was to prevent the system property from being masked.        

However, there is still an issue with the assumption that the working directory is such that a relative path to WEB-INF/logging.properties will correctly resolve.

A fix to the relative path issue has been merged to async-support branch with #150, as the VmRuntimeWebAppContext now looks for a log directory relative to WEB-INF and converts it to an absolute path.

I will close this issue for now - but if you want the fix back ported to master branch prior to the async-support merge, please reopen.

meltsufin:
We need this fixed to be merged into master. However, it looks like it works with Jetty 9.3.5, just not with higher versions.

PR #148, opened 2016-01-05
Fix logging for #136

gregw:
Backport fix from async-support for #136 and #124 

Update System property handling so that logging config is not masked.
Update JettyRunner to set the system property.


ludoch:
LGTM

ludoch:
Will need this to async branch asap.

gregw:
OK, I will merge with async branch now.

PR #152, opened 2016-01-07
Fixed incorrect log path #140

gregw:
Fixed incorrect log path.

The testwebapp starts within the image now (using the testMetadataServer)


ludoch:
Good catch!

PR #150, opened 2016-01-06
Issue 140 rebased to async-support with fixes

gregw:
This is not ready to merge. Created for feedback for #140 against #141

This adds a jetty9-compat-base module, which produces a jetty base artifact which can be run directly with a command like:
```shell
java \
-Droot.webapp=../webapps/testwebapp/ \
-Dcom.google.apphosting.logs=./logs \
-jar ../jetty-distribution-9.3.5.v20151012/start.jar \
--module=testMetadataServer
```

I still need to add support for compiling quickstart.

PR #157, opened 2016-01-13
updated to jetty-9.3.7.RC1

gregw:
Fix for quickstart issue identified in #150


ludoch:
Before we merge, I would like to do a release to maven of the content of this branch. 
Then I can merge tomorrow.
Or what is the timeline for RCs to be GA?

gregw:
We don't have a formal time between RC and final releases, and often don't do RC's for point releases.  We are waiting for one other unrelated issue, which needs to be fixed today or will be in the next release, so we can probably have a 9.3.7.v201601xx before the end of the week.

ludoch:
I would wait, as we would not push an image based on RC code I think.

gregw:
@ludoch updated for openssl debs, which again changed in non durable store.

gregw:
This PR is perhaps no longer needed as #162 updates to the GA version

Issue #158, opened 2016-01-14
Enable logging in jetty9 docker image


gregw:
This will need to use the appengine-java-logging artefact

gregw:
@ludoch Do you want request logging; server logging; or both enabled in the jetty9 image?

gregw:
@ludoch if it is to be server logging as well, then should I rename appengine-java-logging to mvm-java-logging?

ludoch:
Both would be nice. Let's not rename for now, naming is in flux....

Ludo

PR #162, opened 2016-01-14
PR Issue 158

gregw:
Not really ready to merge (at very least it needs some README.md luv'n), but I just wanted to capture work in progress and get some feedback for #158 

This adds a jetty-base to the jetty9 image that has a mvm module that turns on server and request logging.  

This base (and it's module) are removed in jetty9-compat, so nothing is inherited (although perhaps some of it should be?)


gregw:
Ah - just saw your comments about gae rather than mvm
The only issue I have with using gae, is that would mean we'd have 2 different versions of the same name - one under jetty9 and another under jetty9-compat - this could be good or it could be confusing....

But as naming is in flux I will rename to gae pending further review.

gregw:
renaming done

gregw:
@ludoch The GA release is on it's way to maven central now.

gregw:
@ludoch I think this PR is ready to merge now.     I   split the jetty9 module into jetty9-base and jetty9, which is the same structure as jetty9-compat-base and jetty9-compat.  I'm using slightly neater plugins in the pom.xml for jetty9 thanks to some ideas from @joakime (which I think we should use throughout the poms @joakime is preparing a PR)

Also added a GoogleRequestCustomizer to handle GAE proxy headers.

ludoch:
LGTM

Issue #143, opened 2015-12-30
libssl artifact no longer available


gregw:
The master build is currently failing as the libssl1.0.2_1.0.2d-3_amd64.deb is no longer available from 
http://ftp.us.debian.org/debian/pool/main/o/openssl/libssl1.0.2_1.0.2d-3_amd64.deb

A libssl1.0.2_1.0.2e-1_amd64.deb is now available from that location, but we cannot be forced into upgrades like that. We either need to find a durable store for deb, or export one into the maven repo, so it become durable.

joakime:
It seems that the original URL to libssl1.0.2d is for the sid (unstable) tree of Debian.
libssl1.0.2 is not present in in the packaging or indexes for debian stable, debian 8.2, or debian wheezy (those releases are all on libssl1.0.0)

libssl version 1.0.2 is also not present on `archive.debian.org`

libssl1.0.1k is part of jessie(stable) though

The requirement for libssl1.0.2 is actually going to change, frequently, until it enters into the stable trees.

Can you make this requirement multi-stage?

Technique 1:

 1. Get the index from http://ftp.us.debian.org/debian/indices/files/components/arch-amd64.list.gz
 2. Then use the regex `pool\/main\/o\/openssl\/libssl1\.0\.2.*amd64\.deb$` to get the last hit in that index.
 3. That last result is either what you fetch, or a build error?

Technique 2:

 1. Get https://packages.debian.org/sid/amd64/libssl1.0.2/download
 2. Use the link to http://ftp.us.debian.org/ as the "live" version

joakime:
libssl 1.0.2 is also available on `stretch (testing)`

https://packages.debian.org/stretch/libssl1.0.2

joakime:
Link to keep track of all libssl1.0.2 related packages (no matter what suite its found in) - https://packages.debian.org/search?keywords=libssl1.0.2

joakime:
Completed as part of PR #177

PR #177, opened 2016-02-17
Issue #143 - Use (sid) to obtain libssl & openssl

joakime:
- Fix for the constantly changing libssl reference.
  Will temporarily install the (sid) apt repository,
  then install libssl1.0.2 and openssl only,
  removing the sid repository afterwords.

This is a speculative fix, and if you follow the build you'll see that it only adds those dependencies, but it somehow feels dirty to do it this way.

### DIFF
--- a/openjdk8/src/main/docker/Dockerfile
+++ b/openjdk8/src/main/docker/Dockerfile
@@ -35,14 +35,12 @@ ADD sid.list /etc/apt/sources.list.d/
 RUN apt-get -y update \
  && apt-get -y -q --no-install-recommends install \
     libssl1.0.2 \
-    openssl
-
+    openssl \
 # Cleanup sid references
-RUN rm /etc/apt/sources.list.d/sid.list \
- && apt-get -y update
-
+ && rm /etc/apt/sources.list.d/sid.list \
+ && apt-get -y update \
 # Cleanup apt-get temporary files
-RUN apt-get -y -q upgrade \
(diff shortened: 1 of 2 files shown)


joakime:
How do I nudge circleci to build again with no updated commit?  
(The error it got was from a network fetch issue)

```
Failed to fetch http://httpredir.debian.org/debian/dists/jessie-backports/main/binary-amd64/Packages
Hash Sum mismatch
```

ludoch:
There is a rebuild button top right... Let's see what happens when I click
it...

openjdk8/src/main/docker/Dockerfile
```diff
@@ -30,12 +30,19 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
...
+ && apt-get -y -q --no-install-recommends install \
+    libssl1.0.2 \
+    openssl
+
+# Cleanup sid references
+RUN rm /etc/apt/sources.list.d/sid.list \
+ && apt-get -y update
+
+# Cleanup apt-get temporary files
+RUN apt-get -y -q upgrade \
```
ludoch: Could we merge all the runs into a single Docker run step to minimize intermediate layers?

gregw:
mentioning Issue #143 in comment to get cross link to the issue

ludoch:
LGTM

PR #178, opened 2016-02-18
Cleaning up .gitignore

joakime:
Adding / cleaning up the various patterns in your .gitignore to be more useful in the general sense.


ludoch:
LGTM

Issue #164, opened 2016-01-26
ASYNC jetty9-compat image - Incorrect URL generated in a compat image for login


lesv:
https://accounts.google.com/ServiceLogin?
service=ah
&passive=true
&continue=https%3A%2F%2Fappengine.google.com%2F_ah%2Fconflogin%3Fcontinue%3Dhttp%3A%2F%2F0.20160126t110119.lesv-004.appspot.com%2Fguestbook.jsp
&ltmpl=gm
&shdf=ChQLEgZxxxxxxxxxxx

1. Using https://github.com/GoogleCloudPlatform/appengine-java-vm-guestbook-extras/
2. `cd stage3; mvn gcloud:deploy`; visit site - logout, login note URL's.
https://accounts.google.com/ServiceLogin?
service=ah
&passive=true
&continue=https%3A%2F%2Fappengine.google.com%2F_ah%2Fconflogin%3Fcontinue%3Dhttps%3A%2F%2F20160126t111625-dot-lesv-004.appspot.com%2F
&ltmpl=gm
&shdf=ChQLEgZxxxxxxx

Note - **https://20160126t111625.lesv-004.appspot.com/guestbook.jsp**

3. `vi src/main/webapp/Dockerfile`
4. add the suffix **:githubheadasync** to the FROM line
5. `mvn gcloud:deploy`

The Redirect URL on the ASYNC is **http://0.20160126t110119.lesv-004.appspot.com/guestbook.jsp** which is wrong for several reasons:

  a. It's to a specific Instance
  b. It's not using https

The correct redirect URL should be **https://20160126t110119.lesv-004.appspot.com/guestbook.jsp**

lesv:
@ludoch This is IMPORTANT

ludoch:
For Greg, when he is back next week.

ludoch:
Relevant code path: https://code.google.com/p/googleappengine/source/browse/trunk/java/src/main/com/google/appengine/api/users/UserServiceImpl.java

lesv:
If I add a security constraint to the web.xml of CONFIDENTIAL, this problem doesn't occur, the earlier jetty-compat didn't require this.

joakime:
Interesting find about the CONFIDENTIAL bit, I wonder if its related to https://bugs.eclipse.org/486511 (recently identified and fixed in Jetty 9.3.x branches)

ludoch:
haha!

gregw:
[...]

PR #182, opened 2016-02-22
Issue #164

gregw:
Work around for issue #164 by editing the generated URLs.   Tested against guestbook stage 3

### DIFF
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
@@ -167,11 +167,12 @@ public class VmApiProxyDelegate implements ApiProxy.Delegate<VmApiProxyEnvironme
           "content-length: %s. Took %s ms.", packageName, methodName, requestData.length, 200,
           responseData.length, (end - start)));
       
-      // HACK TO FIX USER_SERVICE BUG
+      // HACK TO FIX USER_SERVICE ISSUE #164
       if ("user".equals(packageName)) {
+        boolean userservicefix = !Boolean.getBoolean("UserService164");
         String host = (String) environment.getAttributes().get("com.google.appengine.runtime.host");
         String https = (String) environment.getAttributes().get("com.google.appengine.runtime.https");
-        if (host != null && host.length() > 0) {
+        if (userservicefix && host != null && host.length() > 0) {
           try {
             if ("CreateLogoutURL".equals(methodName)) {
               CreateLogoutURLResponse response = new CreateLogoutURLResponse();
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
@@ -35,6 +35,8 @@ import com.google.apphosting.api.ApiProxy.ApiConfig;
 import com.google.apphosting.api.ApiProxy.ApiProxyException;
 import com.google.apphosting.api.ApiProxy.LogRecord;
 import com.google.apphosting.api.ApiProxy.RPCFailedException;
+import com.google.apphosting.api.UserServicePb.CreateLogoutURLRequest;
+import com.google.apphosting.api.UserServicePb.CreateLogoutURLResponse;
 import com.google.apphosting.utils.remoteapi.RemoteApiPb;
 
 import org.apache.http.HttpResponse;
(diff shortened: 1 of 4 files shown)


gregw:
Ooops forgot to add the environment variable to disable!   Tomorrow...

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
```diff
@@ -159,6 +166,36 @@ public VmApiProxyDelegate() {
           "complete. Service bridge status code: %s; response " +
           "content-length: %s. Took %s ms.", packageName, methodName, requestData.length, 200,
           responseData.length, (end - start)));
+      
+      // HACK TO FIX USER_SERVICE BUG
+      if ("user".equals(packageName)) {
+        String host = (String) environment.getAttributes().get("com.google.appengine.runtime.host");
+        String https = (String) environment.getAttributes().get("com.google.appengine.runtime.https");
+        if (host != null && host.length() > 0) {
```
ludoch: How or where these 2 attributes are set?

gregw: @ludoch I added those attributes to the AttributeMapping enum in VmApiProxyEnvironment

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyEnvironment.java
```diff
@@ -207,7 +205,16 @@
     APPSERVER_TASK_BNS(
         "X-AppEngine-Appserver-Task-Bns",
         "com.google.appengine.runtime.appserver_task_bns",
-        "", true);
+        "", true),
+    HTTPS(
```
gregw: @ludoch the extra attributes are set here

jetty9/pom.xml
```diff
@@ -27,12 +27,6 @@
       <version>${project.version}</version>
       <type>tar.gz</type>
     </dependency>
-    <dependency>
```
gregw: This is an unrelated fix.  Should be in async-support branch anyway to avoid unneccessary transitive dependencies.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyEnvironment.java
ludoch: Right.... But how is setting them? I am not sure they are set in the incoming request sent from the app server bridge (outside the jvm?)

jetty9/pom.xml
gregw: @ludoch The Host header is required for HTTP/1.1 and commonly sent with HTTP/1.0.   The X-AppEngine-Https header is used elsewhere in the code to determine if the request is secure.  You can see both headers in a [dump](https://jetty9-work.appspot.com/dump/info)

gregw: I just tried to hit the site without a Host header and it failed because it is a virtual host sharing an IP address.  So no host header will always result in a failed request.

ludoch: Perfect... remaining issue is the flag to turn this off when we can

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
```diff
@@ -159,6 +166,37 @@ public VmApiProxyDelegate() {
           "complete. Service bridge status code: %s; response " +
           "content-length: %s. Took %s ms.", packageName, methodName, requestData.length, 200,
           responseData.length, (end - start)));
+      
+      // HACK TO FIX USER_SERVICE ISSUE #164
+      if ("user".equals(packageName)) {
+        boolean userservicefix = !Boolean.getBoolean("UserService164");
```
gregw: System prop to disable has been added.  @ludoch maybe a bad name?

ludoch: yes.... Maybe useLocalUserService ?
And true by default

gregw: Well it is not exactly "Local", as we still call the RPC, we just correct the results.   Perhaps UserServiceLocalSchemeHost?

ludoch: ok

ludoch:
LGTM

Issue #189, opened 2016-03-01
Google API's broken


estrellis:
Here is the diagnostic I see:
INFO: HTTP ApiProxy I/O error for logservice.Flush: Network is unreachable

To produce the issue I just used the API in an app

ludoch:
Only for the new environment (not impacting customers yet)
Need to apply a fix we did internally. Simple change.

PR #190, opened 2016-03-01
Apis usage using the new environment.

ludoch:
Bug https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/issues/189


lesv:
Shouldn't this figure out which environment you are in before choosing that endpoint?

ludoch:
It works for both environments.

Issue #201, opened 2016-03-16
Compare head and async branch regarding workload


ludoch:
We did some informal load test of a web app using GAE APIs and it seems the async-branch delivers lower QPS than head...
It might be due to either the async work done, or the global API ticket usage.

I will provide the test web app (it will be in Open Source soon) and the way to run the load test.
We need to reproduce and analyse the diff with different configurations.
We can then start a regular perf end2end test that we could run nightly on our Jenkins far to track possible regressions early in the dev cycle.

gregw:
first quick look at benchmark tool indicates that it is trying to measure the right thing.    Tools that report "achieved" throughput are rarely actually measuring throughput.   This tool is correctly offering load in increments and charting the achieved latencies at 90% 95% and 99% to see if an acceptable QoS is achieved.       So that is good.... now have to look into why async branch is doing badly.

gregw:
OK, I take it back about the wrk benchmark tool.  The load it offers is a function of observed latency, so it's throughput numbers are actually measuring latency rather than throughput.  

However, having said that, I'm seeing 100ms more latency on the async branch than from master, even on a non-authenticated path that should do no API calls.    That is a long time, so I need to work out how to work out what is going on.....  stand by...

gregw:
progress report.... I've been deploying instrumented versions of the async image and there is an unusual latency spent between dispatching the request and entering the context, which suggest the session manager.    I will zero in a bit more today.

gregw:
History logged on [instrumented image](https://20160322t120513-dot-jetty9-work.appspot.com/dump/) definitely indicates that a lot of time is spent in session manager doScope:
````
92361917790408, 92361917790408,Endpoint Opened
    54280, 92361917844688,Last Selected
   180932, 92361918025620,RequestContext created
    41280, 92361918066900,VmApiProxyEnv created
     3643, 92361918070543,enterScope
     4504, 92361918075047,session.doScope
 20541498, 92361938616545,servlet.doScope
    23490, 92361938640035,requestInitialized
     7655, 92361938647690,session.doHandle
     1728, 92361938649418,security.handle
    62764, 92361938712182,servlet.doHandle
    16212, 92361938728394,doGet
   591225, 92361939319619,response written
   188396, 92361939508015,flushed
````

gregw:
A bit more instrumentation shows that the vast majority of the time is spent loading the session:
````
367386447442, 367386447442,Endpoint Opened
[...]

PR #205, opened 2016-04-01
Issue #201 User request ticket instead of global ticket before commit

gregw:
This is not yet ready to merge as it needs testing.  PR created now to facilitate discussion on #201

### DIFF
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyEnvironment.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyEnvironment.java
@@ -529,12 +529,12 @@ public class VmApiProxyEnvironment implements ApiProxy.Environment {
   }
 
   public void clearTicket() {
-    ticket=null;
+    ticket = null;
   }
   
   public String getTicket() {
-    String t=ticket;
-    return (t!=null&&!t.isEmpty())?t:globalTicket;
+    String t = ticket;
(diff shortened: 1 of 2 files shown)


appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyEnvironment.java
```diff
@@ -526,8 +528,13 @@ public String getServer() {
     return server;
   }
 
+  public void clearTicket() {
+    ticket=null;
+  }
+  
   public String getTicket() {
-    return ticket;
+    String t=ticket;
```
ludoch: nit: space before and after =

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyEnvironment.java
```diff
@@ -526,8 +528,13 @@ public String getServer() {
...
+    return (t!=null&&!t.isEmpty())?t:globalTicket;
   }
```
ludoch: space before and after operators

gregw:
I signed it!

ludoch:
Ready for merge or doing more testing?

gregw:
Ludo,

I think the changes in this PR are good.   It is certainly better than just using global ticket, but we should probably open an issue somewhere to clarify how long the request ticket is valid for, so we can keep using it as long as possible.

gregw:
... so yes good to merge

PR #209, opened 2016-04-07
Fix #186 to use correct webdefault.xml

### LINKED ISSUE #186: Session content not persisted to datastore with Managed VMs
See https://code.google.com/p/googleappengine/issues/detail?id=12766

Maybe some configuration issue?

janbartel:
Fix for  #186. root.xml was referring to etc/webdefault.xml in jetty.home instead of jetty.base.


janbartel:
I signed it!

ludoch:
Nice. Thanks! I'll file a bug to make sure we can test this going forward.

PR #211, opened 2016-04-12
Create $JETTY_BASE/work so that Jetty explodes WARs there

chanseokoh:
As per internal discussion. Bug number 27900546.

### DIFF
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -36,6 +36,7 @@ ENV JETTY_BASE /var/lib/jetty
 ADD jetty-base $JETTY_BASE
 WORKDIR $JETTY_BASE
 RUN java -jar $JETTY_HOME/start.jar --add-to-startd=setuid \
+ && mkdir $JETTY_BASE/work \
  && chown -R jetty:jetty $JETTY_BASE \
  && mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty \
  && ln -s $JETTY_BASE/webapps/root /app
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -36,7 +36,8 @@ ENV JETTY_BASE /var/lib/jetty
 ADD jetty-base $JETTY_BASE
 WORKDIR $JETTY_BASE
 RUN java -jar $JETTY_HOME/start.jar --add-to-startd=setuid \
- && mkdir $JETTY_BASE/work \
+ && mkdir $JETTY_BASE/work
+ && ln -s $JETTY_BASE/work/jetty-0.0.0.0-8080-root.war-_-any- $JETTY_BASE/root-app
  && chown -R jetty:jetty $JETTY_BASE \
  && mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty \
  && ln -s $JETTY_BASE/webapps/root /app


chanseokoh:
FYI, this is for the async-support branch.

ludoch:
Let me add your comments from the internal tool:

"There can be many other alternatives (possibly better ones), but I find the following method is one of the quickest solution.
* Proposed solution
By creating a directory $JETTY_BASE/work (/var/lib/jetty/work), I checked that Jetty9 in our image always creates the following directory and explodes a WAR into it:

$JETTY_BASE/work/jetty-0.0.0.0-8080-root.war-_-any-/

(Random numbers don't appear in the directory name unlike when Jetty explodes a WAR into a tmp directory. I hope this is not a bug in Jetty but rather intended). Then, Java class files will be placed in

$JETTY_BASE/work/jetty-0.0.0.0-8080-root.war-_-any-/webapp/WEB-INF/classes

This of course assumes that we use a WAR named 'root.war', which we currently do.

ludoch:
And the cloud debugger will be changed to could to search for the above directory

ludoch:
for @gregw  to validate the jetty-base/work location as stable for exploded war.

joakime:
Is the purpose of this is to establish a known directory name for the exploded war?

chanseokoh:
@joakime, Yes. If there is a better way to make Jetty reliably and consistently explode a war into a fixed directory, that will be better. Currently, Jetty explodes a war under something like '/tmp/jetty-0.0.0.0-8080-root.war-_-any-<long-random-number>.dir'.

chanseokoh:
I made another change to create a symlink for the long directory name. This should please the Cloud Debugger team for hard-coding this specific path in their source code.

joakime:
You can use the `WebAppContext.setTempDirectory()` to fix the temp directory to anything you need/want.

Example:
```xml
<?xml version="1.0"?>
<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_9_3.dtd">
<Configure id="webapp" class="org.eclipse.jetty.webapp.WebAppContext">
  <Set name="contextPath">/root</Set>
  <Set name="war"><Property name="jetty.webapps" default="."/>/root.war</Set>
  <Set name="extractWAR">true</Set>
  <Set name="tempDirectory"><Property name="jetty.base" default="."/>/work/root-exploded</Set>
</Configure>
```

The above file `${jetty.base}/webapps/root.xml` will set the directory to be `${jetty.base}/work/root-exploded`.

You'll see something along these lines in your logging ...

```
2016-04-12 14:11:22.283:INFO:oejsh.ContextHandler:main: Started
  o.e.j.w.WebAppContext@2d3fcdbd{/root,file:///home/joakim/distros/
  tempdir-base/work/root-exploded/webapp/,AVAILABLE}{/root.war}
```

chanseokoh:
I tested that method inside our image a few days ago. In my test, Jetty created the directory `${jetty.base}/work/root-exploded`, but it was basically empty. Strangely, it nevertheless created the usual `/tmp/jetty-0.0.0.0-8080-root.war-_-any-<long random number>.dir`. I will test this again, but could you also check it on your side that it doesn't create a directory under `/tmp` and `${jetty.base}/work/root-exploded` is not empty (it may contain one or two sub-directories though).

chanseokoh:
Ah, seems like it's working. I should have done something wrong.

I will shortly create another PR to include this 'root.xml' in our images.

gregw:
@chanseokoh I'm not a big fan of the symlink to the directory within work as depends on jetty never changing it's name generation.

A work directory is used primarily when generating/compiling JSPs so that the results can be results can be re-used if the application is stopped/restarted.

If there are not going to be multiple contexts and if they are not going to be stopped and restarted, then I think the approach of just setting the temporary directory to a known location is simpler and preferable.

chanseokoh:
@gregw Please see #213. This is closed. No symlink anymore. #213 does set a temp directory to a known location.

PR #212, opened 2016-04-12
Create $JETTY_BASE/work so that Jetty explodes WARs there

chanseokoh:
As per internal discussion. Bug number 27900546.


chanseokoh:
Per #211, I will create 'root.xml' in another PR. Thanks!

PR #213, opened 2016-04-12
Explode WAR into $JETTY_BASE/webapps/root

chanseokoh:
Following the discussion in #211, we make Jetty explode `root.war` into `$JETTY_BASE/webapps/root-exploded` in a proper way.

### DIFF
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -40,7 +40,7 @@ RUN java -jar $JETTY_HOME/start.jar --add-to-startd=setuid \
  && chown -R jetty:jetty $JETTY_BASE \
  && mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty \
  && ln -s $JETTY_BASE/webapps/root /app \
- && touch $JETTY_BASE/webapps/root.xml
+ && touch $JETTY_BASE/webapps/root.war
 
 # Start Jetty directly
 COPY docker-entrypoint.bash /
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -36,9 +36,11 @@ ENV JETTY_BASE /var/lib/jetty
 ADD jetty-base $JETTY_BASE
 WORKDIR $JETTY_BASE
 RUN java -jar $JETTY_HOME/start.jar --add-to-startd=setuid \
+ && mkdir $JETTY_BASE/webapps/root \
  && chown -R jetty:jetty $JETTY_BASE \
  && mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty \
- && ln -s $JETTY_BASE/webapps/root /app
+ && ln -s $JETTY_BASE/webapps/root /app \
+ && touch $JETTY_BASE/webapps/root.xml
 
 # Start Jetty directly
 COPY docker-entrypoint.bash /
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -36,7 +36,7 @@ ENV JETTY_BASE /var/lib/jetty
 ADD jetty-base $JETTY_BASE
 WORKDIR $JETTY_BASE
 RUN java -jar $JETTY_HOME/start.jar --add-to-startd=setuid \
- && mkdir $JETTY_BASE/webapps/root \
+ && mkdir -p $JETTY_BASE/webapps/root \
  && chown -R jetty:jetty $JETTY_BASE \
  && mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty \
  && ln -s $JETTY_BASE/webapps/root /app \
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -40,7 +40,7 @@ RUN java -jar $JETTY_HOME/start.jar --add-to-startd=setuid \
  && chown -R jetty:jetty $JETTY_BASE \
  && mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty \
  && ln -s $JETTY_BASE/webapps/root /app \
- && touch $JETTY_BASE/webapps/root.war
+ && touch --date '1970/01/01' $JETTY_BASE/webapps/root
 
 # Start Jetty directly
 COPY docker-entrypoint.bash /


chanseokoh:
Just a heads-up for the next step:

I created a CL to be reviewed by the Cloud Debugger team to make a debugger agent to look for `$JETTY_BASE/webapps/root-exploded/webapp/WEB-INF/classes` and `$JETTY_BASE/webapps/root-exploded/webapp/WEB-INF/lib`.

Once they approve the CL and the changes are pushed into prod, it will be available for [fetch by Dockerfile for the openjdk image](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/master/gke-debian-openjdk/src/main/docker/Dockerfile#L47). I'll let you know after I confirm that the CL is pushed to prod. Then, please rebuild and publish all the images to public.

jetty9/src/main/docker/root.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_9_3.dtd">
+<Configure id="webapp" class="org.eclipse.jetty.webapp.WebAppContext">
+  <Set name="contextPath">/root</Set>
+  <Set name="war"><Property name="jetty.webapps" default="."/>/root.war</Set>
+  <Set name="extractWAR">true</Set>
+  <Set name="tempDirectory"><Property name="jetty.base" default="."/>/webapps/root-exploded</Set>
```
gregw: the standard for in situ extracting is to expand to webapps/root/ 
Having a temporary directory is a separate matter, although temporary extraction is done to the temporary directory.

gregw: So I checked the exact logic on the jetty unpacker and to unpack from webapps/root.war to webapps/root/   you need to set extractWAR to true (done) AND the timestamp of the war file needs to be more recent than the timestamp of the directory.  So I think you can probably achieve what you want with:
````
mkdir webapps/root
touch --date 1970/01/01 webapps/test
````

Or better yet, just arrange for the root.war to be unpacked by a Dockerfile, so it happens as the image is built rather than when the image is run.

Comment (chanseokoh) [inline jetty9/src/main/docker/root.xml]: @gregw Unpacking root.war by a Dockerfile does not seem to be a good solution, because Jetty explodes root.war with its own custom directory structure, which I think can change in later Jetty versions.

I was able to make Jetty unpack root.war into any directory other than `webapps/root`, but had trouble unpack into `webapps/root`. I will try your method, and if it works, I will use the `touch` method.

**Comment by chanseokoh** [comment]
@gregw I checked that `touch --date 1970/01/01` makes Jetty explode root.war into `webapps/root`. I have updated the Dockerfile and root.xml accordingly. PTAL. Please also test if it really works on the base image itself.

**Comment by gregw** [comment]
LGTM.   However, note that when making the final image, if the root.war can be exploded into root/  then startup will be much quicker.   This configuration will not need to change as Jetty will detect if root/ is newer than root.war and will not unpack during startup.

For these PR, should I assign to @ludoch or @aozarov now to do the merge? Or shall I?

**Comment by chanseokoh** [comment]
@ludoch said he will be out of office until Apr 25. I have no idea who should do the merge. I think you can decide it best.

About manually exploding root.war into root/ before Jetty starts up: I tried this before, and also just tried it again. No matter what I do (e.g., using root.xml or having root.war besides root/), what happens is that Jetty doesn't recognize the things inside root/ as an web app. So, it ends up listing directories and files on a browser natively as if they are regular files. For this reason, I think it is ideal to make Jetty recognize and explode a war.

**Comment by chanseokoh** [comment]
Ah, could you confirm that Jetty really explodes a war into root/ in the final images (openjdk8 and jetty9)? I've only tested it locally doing things on top of the base image.

**Comment by gregw** [comment]
OK then once we are happy with this, I'm OK to do the merge as it currently is a very small change in behaviour, it still deploys root.war, but just changes where it is unpacked.... 

However, I do now have some concerns.  The link to /app looks like it was already assuming the war file would be unpacked to root/ by something.

Your unpacking to root/ should have worked (with or without these changes), so something strange is going on that we need to understand.        I'll confirm the behaviour of the current image later today and get back to you.

**Comment by chanseokoh** [comment]
Thanks!

You will certainly see that currently, root.war is not exploded into root/ but into a temp directory. This behavior is well documented [here](http://www.eclipse.org/jetty/documentation/current/ref-temporary-directories.html). So, root/ is empty anyway. My guess is that someone mistakenly assumed that Jetty will use root/ based on behaviors of previous Jetty versions, but who knows?

**Comment by chanseokoh** [comment]
I was doing some testing, and it turns out that root.xml is not necessary. It looks like Jetty should have exploded root.war into root/, as you said. However, I think the problem was that root/ was newer than root.war. To summarize, in order to make Jetty explode root.war into root/:

1. root/ must exist before Jetty starts.
2. root.war should be newer than root/.
3. root.xml is not necessary.

So, a better solution is to just touch the root.war. I think it is better than touching root/ with --date 1970/01/01. I'll update this PR accordingly.

**Comment by chanseokoh** [comment]
Now we just touch root.war. Please look into the final image to make sure that nothing touches root/ later, and root.war is indeed exploded into root/. Thanks!

**Comment by gregw** [comment]
@chanseokoh actually a typical Dockerfile for using these images looks like:
````docker
FROM jetty9:latest
ADD . /app
````
Which indicates that the builder is expecting the current directory to contain an already exploded webapplication, which is added to /app, which has previously been linked to $JETTY_BASE/webapps/root

So if you put a root.war file in the docker base directory, you will end up with `webapps/root/root.war` which will not be deployed as a webapp, only a directory containing a single file.    If you explode the war into a subdirectory of the docker base called root, then you will get `webapps/root/root/` which will also not deploy as a webapp.

So I think it is definitely intended that the docker base directory contain the exploded war file.  I've just retested this and that is indeed how it works.

I guess with this alternate configuration, you can put root.war into webapps and jetty will unpack it to /apps.   But I think that is a waste of resources to repeatedly unpack every time an instance is started

**Comment by chanseokoh** [comment]
@gregw I see. That typical use case makes sense. Unfortunately, our gcloud CLI tool auto-generates a Dockerfile as follows:

```
FROM gcr.io/google_appengine/jetty9
ADD target.war $JETTY_BASE/webapps/root.war
```

We don't have control over the gcloud tool and their decision to produce this Dockerfile. We need to make sure this Dockerfile works with App Engine flexible environment.

**Comment by chanseokoh** [comment]
Oh, to clarify, the above Dockerfile currently works with flex. It's just that the Cloud Debugger is broken. Currently, the Debugger is configured to look into webapps/root/WEB-INF/{classes,lib}. So, we do need to explode a war into root/.

**Comment by chanseokoh** [comment]
Now you said it, I think it is not right to touch `root.war`, which may not exist in typical use cases. I'll update this PR tomorrow. So then, do you think that the right solution is to touch the `root/` directory with `--date 1970/01/01`? Do you think it will work on any use cases?

**Comment by gregw** [comment]
Currently the images are setup to expect the unpacked webapp to be added to `/app`, which is setup as a symlink to `${JETTY_BASE}/webapps/root/`

I think that is the preferred option.

touching root.war is definitely wrong, not least as you lose the date the war file was created, which can be useful for debugging.

Creating the empty `${JETTY_BASE}/webapps/root/` directory and touching to 1970 may allow for either the unpack app to be added to `/app` OR the packed war being added to `${JETTY_BASE}/webapps/root.war`, but I'd like to test it a bit to ensure that this does not break anything

**Comment by chanseokoh** [comment]
Thanks. I reverted the Dockerfile to use `touch --date '1970/01/01'` for now.

**Comment by chanseokoh** [comment]
@gregw How's the testing going?

**Comment by gregw** [comment]
@chanseokoh I'm not meant to be working on this until tomorrow, but will try to squeeze some testing in today.   It's been a bit chaotic, so want to get a clear picture of the current state and impacts before a merge.

**Comment by gregw** [comment]
So testing is good (well this is a pretty minor change now).

Building an image with an unpacked war can use a Dockerfile like:
````docker
FROM jetty9:latest
ADD . /app
````

If you want to build and image with a war in it, you need to update an ENV var:
````docker
FROM jetty9:latest
ENV root.webapp ./webapps/root.war
ADD root.war $JETTY_BASE/webapps/root.war
````
Note that this still does the unpack into webapps/root/ which can be seen in the log output:
````
Apr 18, 2016 11:16:10 PM org.eclipse.jetty.server.handler.ContextHandler doStart
INFO: Started o.e.j.w.WebAppContext@7823a2f9{/,file:///var/lib/jetty/webapps/root/,AVAILABLE}{/root.war}
````
which shows the docroot as `/root/` for the `/root.war` webapp

So I'm good to merge this into async branch.   Will look at the test failures in master tomorrow.

**Comment by ludoch** [comment]
Hi,
this PR will not work with the fingerprinting done in the Cloud SDK which generates the Dockefile for the different images.
The Java PRD says that for the pure Jetty9 image, only a .war file is used, and does not define this ENV variable...
We need to rollback for now, and I'll sync up with the Cloud debugger folks when I am back.
Or the unzip should be done in the Dockefile from the war file added to it?
You can see the fingerprinting logic in the cloud sdk lib/googlecloudsdk/api_lib/app/runtimes/java.py
and platform/ext-runtime/java/bin/generate_configs files

**Comment by chanseokoh** [comment]
Thanks, @gregw.

What's with the new ENV var? There's no code change related to the var in this PR. Also, I thought unpacking root.war will work fine with the following Dockerfile, which is what current fingerprinting generates:

```
FROM gcr.io/google_appengine/jetty9
ADD <some valid .war> $JETTY_BASE/webapps/root.war
```

At least it seems to work on my local environment. Could you confirm this? If the current merge works as expected, then we won't need to rollback it.

chanseokoh:
Also, the docroot should be `/` but not `/root`. On my local environment, the docroot is `/`. Probably it is also the case in the images, but please double-check that.

chanseokoh:
Ah, sorry I think I misunderstood the meaning of the docroot. As long as the contents in the docroot can be reached directly via the `/` path in a URL, it's fine.

gregw:
@ludoch Ah yes the ENV variable root.webapp is not needed for the jetty9 image as there is no root.xml file that needs to be told to use `webapps/root/` or `webapps/root.war`.  Anything added to `webapps` will be deployed using the normal jetty deployer rules.

So prior to this change,  a symlink is made from  `/app`, `webapps/root/` which will initially be a broken link (and hence part of the problem for the debugger).   If the developer happens to add a webapp that arrives in `webapps/root/`, then it will be visible in `/app`, however the chances of that are not great as a `root.war` will probably be unpacked to a temporary location rather than `webapps/root/`

With this change, I think the situation is a little bit better because the `webapps/root/` directory is created as an empty directory and thus the `/app` link is not broken.        A developer can then deploy in several different ways that all appear to work:

- An unpacked war can be ADDED to `/app`
- An unpacked war can be ADDED to `webapps/root/`
- A packed war can bee ADDED to `webapps/root.war`

In all these cases I believe that the root webapp is correctly deployed and it's contents will be visible in /app.   So the only deployment mode that is a little broken by this is if only non-root webapps are deployed, as this will then end up with an empty directory as the / webapp.

If that is an important deployment mode, then I think we should roll back.     But if it is not, then I don't think this merge is harmful.... but I agree we've got ourselves very confused by it all and need to stop and think some more before our next step.

chanseokoh:
@gregw Thanks for the explanation. Hope this is settled down soon.

To summarize and re-iterate, what I want is actually very trivial:

1. Should work with the current Dockerfile (that I posted above), as always has been. `root.war` will be added to `webapps/root.war`.
2. `root.war` should be exploded into `webapps/root/` (so that Stackdriver Debugger can see `webapps/root/WEB-INF`)

Note that we don't want to change the above two contracts. If any of the above contracts has to change, it will end up in an endless line of discussions among various parties, teams, and PMs. They are also the currently active contracts, and breaking them will need extensive testing across multiple teams.

gregw:
@chanseokoh I believe that your current PR as merged achieves both your contracts.   So I believe it is good to leave merged.    I'll look at your master PR  #214 now.

chanseokoh:
@gregw Thanks for the merge. I verified that the changes are in the 0.97 tag. Frustratingly, ** `root.war` is being not exploded into `webapps/root`.** As such, the new image merely displays an empty directory listing of the `webapps/root` content (i.e., outputing `Directory: /`).

The problem is that `root.war` is owned by the root user. This didn't seem to be a problem when we did not have the `webapps/root` directory, as long as the jetty user have a read access on it (which has been the case). However, when we have the `webapps/root` directory, it seems like the `root.war` should be owned by jetty. Running `chown jetty.jetty $JETTY_BASE/webapps/root.war` solves this problem.

To achieve this, I think we can add the `chown` command in `docker-entrypoint.bash`.

chanseokoh:
Sorry, I was mistaken, so I deleted my previous comment entirely.

I noticed that, **in production**, `$JETTY_BASE/webapps/root` is either cleared, modified, or touched later by something. The timestamp of the directory is later than that of `root.war`. Strangely, I noticed that this doesn't happen when I run the image locally with `docker run <image>`.

I think **in production**, Jetty explodes the `root.war` but somehow everything in `$JETTY_BASE/webapps/root` is deleted later.

@gregw Could you look into this?

chanseokoh:
@gregw Never mind my last comment. I created another PR #222 that will solve this problem. Please take a look at #222.

PR #218, opened 2016-04-14
Fixed #217 NPE

### LINKED ISSUE #217: Async branch throws NPE on context start
The ContextListener in the async branch is not protected for a null request, which can occur with start/stop scope calls.

gregw:
Fixed #217 NPE

### DIFF
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
@@ -341,6 +341,8 @@ public class VmRuntimeWebAppContext
   }
 
   public RequestContext getRequestContext(Request baseRequest) {
+    if (baseRequest == null)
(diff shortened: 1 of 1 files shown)


gregw:
I signed it! already!

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -341,6 +341,8 @@ public boolean isTrustedRemoteAddr(String remoteAddr) {
   }
 
   public RequestContext getRequestContext(Request baseRequest) {
+    if (baseRequest == null)
```
aozarov: I suggest using the Google Style guide in new code and applying the formatter in seperate PRs.
This should be:
```java
if (baseRequest == null) {
  return null;
}
```
but fine as is if you plan to run the formatter in the other PR after pulling this change.

aozarov:
Merging, let's fix the formatting issue in the other (rebased) PR.

PR #219, opened 2016-04-14
Enable Cloud (Stackdriver) Debugger on Java flex by default

chanseokoh:
The Cloud Debugger team confirmed that the Debugger issue that forced us to disable it temporarily on the flex env (including Phython runtimes) has been resolved. The team asked me to take an action to enable it. Note that the Debugger is already enabled by default on the standard env (including jetty9-compat), and it should be automatically enabled for Java flex as documented [here](https://cloud.google.com/debugger/docs/setting-up-java-on-app-engine).

I took the liberty to make a change to setup-env.bash at my discretion, but feel free to jump in.

The document [here](https://cloud.google.com/appengine/docs/flexible/java/dev-jetty9?hl=pt-BR#environment_variables) should also be revised to correct the explanation for the DBG_ENALBE env variable.


chanseokoh:
Internal bug number: 28192592

gregw:
LGTM

chanseokoh:
@gregw. Thanks for the update. Actually, @ludoch is concerned about security and other issues about having it on by default. I guess we will have some internal discussions for a while.

chanseokoh:
Okay! It turns out @ludoch mistook Stackdriver Debugger for a local debugger. We're good to go.

ludoch:
Yes, Greg, if you have fine as well, integrate.

chanseokoh:
Also changed Cloud Debugger to Stackdriver Debugger. It's official.

PR #220, opened 2016-04-14
Enable Stackdriver Debugger by default

chanseokoh:
Do the same thing in #219 for the master branch.

### DIFF
--- a/gke-debian-openjdk/README.md
+++ b/gke-debian-openjdk/README.md
@@ -38,13 +38,13 @@ The entry point for the gke-debian-openjdk image is [docker-entrypoint.bash](htt
 
 If the default command (java) is used, then the entry point sources the [gke-env.bash](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/master/gke-debian-openjdk/src/main/docker/gke-env.bash), which looks for supported features: ALPN, Cloud Debugger & Cloud Profiler.  Each of these features must be explicitly enabled and not disable by environment variables, and each has a script that is run to determine the required JVM arguments:
 
-| Feature        | directory    | Enable        | Disable        | JVM args      |
-|----------------|--------------|---------------|----------------|---------------|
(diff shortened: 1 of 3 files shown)


gke-debian-openjdk/README.md
```diff
@@ -38,13 +38,13 @@ The entry point for the gke-debian-openjdk image is [docker-entrypoint.bash](htt
...
-|----------------|--------------|---------------|----------------|---------------|
-| ALPN           | /opt/alpn/   | $ALPN_ENABLE  | $ALPN_DISABLE  | $ALPN_BOOT    |
-| Cloud Debugger | /opt/cdbg/   | $DBG_ENABLE   | $CDBG_DISABLE  | $DBG_AGENT    |
-| Cloud Profile  | /opt/cprof/  | $CPROF_ENABLE | $CPROF_DISABLE | $PROF_AGENT   |
-| Temporary file |              | $TMPDIR       |                | $SET_TMP      |
-| Java options   |              | $JAVA_OPTS    |                | $JAVA_OPTS    |
+| Feature        | directory    | Enable            | Disable        | JVM args      |
+|----------------|--------------|-------------------|----------------|---------------|
+| ALPN           | /opt/alpn/   | $ALPN_ENABLE      | $ALPN_DISABLE  | $ALPN_BOOT    |
+| Cloud Debugger | /opt/cdbg/   | \<on by default\> | $CDBG_DISABLE  | $DBG_AGENT    |
```
ludoch: Is is a good idea to enable a debugger by default? 
I do not think so. Do we have security team review/approval?

ludoch:
Does the debugger work on AWS as well or it is GCE centric?

ludoch:
Images pushed with tag 0.97

Can you  tried? Just tried a compat runtime with this dockerfile, but could not see the debugger. Maybe it is my setting (no source code yet).

FROM gcr.io/google_appengine/jetty9-compat:0.97

ADD . /app

Once validated, I can make the 0.97 the latest on the 3 images.

chanseokoh:
The changes are in, but I noticed it doesn't work in production. I will follow up this in #213.

PR #222, opened 2016-04-24
Explicitly unpack root.war (if present) into webapps/root

chanseokoh:
The previous PR had two problems:

1) Mysteriously, `webapps/root` is still empty (despite its timestamp is changed from 1970/01/01), which only happens in production but not in a local env.
2) Unpacked contents of `root.war` should be present **before** Jetty starts. If Jetty explodes `root.war` while it is being initialized, Stackdriver Debugger can't load the class and lib files of `root.war`.

So, I suggest the following changes that explicitly unpack `root.war`, as an ultimate solution. It works on my local machine and my own ad-hoc testing in production.

What this PR does:

- Revert the last Pull Requst that creates and touches `webapps/root`. (No need to create the directory since `unzip` will create it anyway.)
- Install `unzip`.
- Unpack `root.war` explicitly if present.
- Set `$RUNTIME_DIR` and pass it to `cdbg/format-env-appengine-vm.sh`. (This env variable that the shell script uses was missing.)

### DIFF
--- a/gke-debian-openjdk/src/main/docker/Dockerfile
+++ b/gke-debian-openjdk/src/main/docker/Dockerfile
@@ -24,7 +24,7 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
     openjdk-8-jre \
     netbase \
     wget \ 
-    zip \
+    unzip \
  && apt-get clean \
  && rm /var/lib/apt/lists/*_*
 
--- a/gke-debian-openjdk/src/main/docker/Dockerfile
+++ b/gke-debian-openjdk/src/main/docker/Dockerfile
@@ -24,6 +24,7 @@ RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/a
     openjdk-8-jre \
     netbase \
     wget \ 
+    zip \
  && apt-get clean \
  && rm /var/lib/apt/lists/*_*
 
--- a/gke-debian-openjdk/src/main/docker/gke-env.bash
+++ b/gke-debian-openjdk/src/main/docker/gke-env.bash
@@ -4,7 +4,7 @@ if [[ -n "$ALPN_ENABLE" ]]; then
   ALPN_BOOT="$( /opt/alpn/format-env-appengine-vm.sh )"
 fi
 
-DBG_AGENT="$( /opt/cdbg/format-env-appengine-vm.sh )"
+DBG_AGENT="$( RUNTIME_DIR=$JETTY_BASE /opt/cdbg/format-env-appengine-vm.sh )"
 if [[ -n "$DBG_ENABLE" ]]; then
   if [[ "$GAE_PARTITION" = "dev" ]]; then
     echo "Running locally and DBG_ENABLE is set, enabling standard Java debugger agent"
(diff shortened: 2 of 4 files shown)


gregw:
@chanseokoh I'm still not clear why you wish the developer to build an image with a root.war and then to unpack that every time the image is started.   A lot of the effort in this project has been to minimize start time and wasted cycles, so this appears to be counter to that effort.

Note that the [README.md](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/tree/master/gke-jetty) for the pure jetty image indicates that the derived images should be built with:
````docker
  FROM YOUR_BUILT_IMAGE
  add . /app
````

As the pure jetty image is only a recent addition to the project,  I can't see how there could be a lot of dependencies wanting to deploy packed wars?  But then I've got a limited view of the greater project..

There are also issues that images built with packed wars are themselves hard to use as the base for derived images that might with to modify the deployed webapp.

Thus I'd really like to triple check that deploying packed wars is something that is both desirable and the preferred deployment model.

If so, then at the very least, if packed war deployment is the preferred mode, this PR needs to update the README.md to describe how to deploy a `root.war` file.  It would also be good to have a date check in the script, to ensure that an older war does not override a newer root directory.

chanseokoh:
@gregw There is an ongoing internal discussion. If things are finally settled, I will modify the unpacking logic so that it does not overwrite existing contents of `webapps/root`, if any.

ludoch:
This is deeply related to the fingerprinting in the Cloud SDK that generates the Dockerfile.
It was decided (PRD) that for pure Jetty9 (not compat mode), we would by default accept a .war file (not exploded) because it is standard, and similar to the .jar file we accept for the openjdk8 image...
So to mimimize Cloud SDK change, we should try to keep accepting .war file.
The internal Dockerfile logic can do whatever is needed for the Cloud Debugger needs.

chanseokoh:
@gregw I made changes so that `root.war` is unpacked only when `webapps/root` does not exist or the directory is completely empty. Comparing timestamps will not work because of [the issue 1) in my first comment](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/pull/222#issue-150690874). (For some unknown reason, in production, the timestamp of `webapps/root` is later than that of `root.war` while the directory is empty.)

I did my own testing [with this method](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/pull/223#issuecomment-214090686). I will be able to do a final testing when a new image is built and released with a temporary tag.

Regarding updating docs (e.g., this [README.md](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/tree/master/gke-jetty) and this [README.md](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime)), I think it is better to create another PR. I'll leave it up to you.

Thanks!

gregw:
LGTM - merged!
standing by for README.md PR.

ludoch:
pushed images tag 0.98 for testing of the master branch
(and githubheadasync tag for the async branch)

gregw:
changes have been merged through to async-support branch

chanseokoh:
I tested both `0.98` and `githubheadasync`. They work seamlessly.

BTW, I was looking at the Java compat image because it also makes use of `RUNTIME_DIR` in its `Dockerfile` and `jetty_run.sh`. The variable in Java compat don't seem necessary anymore after merging this PR. We should probably remove it in that image.

chanseokoh:
Let's publish the `0.98` and `githubheadasync` if you don't see any problem.

chanseokoh:
@ludoch When do you plan to make `0.98` latest? We are ready to roll out, aren't we?

PR #223, opened 2016-04-24
Explicitly unpack root.war (if present) into webapps/root

chanseokoh:
Do the same thing in #222 for the `async-support` branch.

### DIFF
--- a/jetty9/src/main/docker/Dockerfile
+++ b/jetty9/src/main/docker/Dockerfile
@@ -40,7 +40,7 @@ RUN java -jar $JETTY_HOME/start.jar --add-to-startd=setuid \
  && chown -R jetty:jetty $JETTY_BASE \
  && mkdir /tmp/jetty && chown -R jetty:jetty /tmp/jetty \
  && ln -s $JETTY_BASE/webapps/root /app \
- && touch $JETTY_BASE/webapps/root.xml
+ && touch $JETTY_BASE/webapps/root.war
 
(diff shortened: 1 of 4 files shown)


ludoch:
How did you test it live? IE with which published image and custom
dockerfile?

gregw:
Closing this one and we can merge through the change #222 from master to async-branch.  In general it's best to only have a single PR

chanseokoh:
@ludoch I tested this on both the latest and the 0.97 images for the Java8/Jetty9 Runtime. The way I tested this is, basically, manually overwriting `gke-env.bash` and `docker-entrypoint.bash` with the `Dockerfile` below (while trying to retain the structure of the original `Dockerfile` generated by the fingerprinting code). I ran `gcloud preview app deploy` in a directory where there is `target.war` (a HelloWorld app). This successfully deploys the app to App Engine flex, and I can use Stackdriver Debugger on Pantheon.

app.yaml:
```
runtime: custom
vm: true
```

Dockerfile:
```
FROM gcr.io/google_appengine/jetty9
RUN apt-get -y -q --no-install-recommends install unzip
ADD target.war $JETTY_BASE/webapps/root.war
COPY gke-env.bash /
COPY docker-entrypoint.bash /
```

Note that the original `Dockerfile` is
```
FROM gcr.io/google_appengine/jetty9
ADD target.war $JETTY_BASE/webapps/root.war
```

PR #224, opened 2016-04-24
Async support Merged from master

gregw:
This PR contains no changes, but just updates the sync point from master to async-support, so that future changes to master can easily be merged into async-support.    It is essentially a catchup merge that cleans up changes that were made in parallel to both branches.


gregw:
I signed it!

PR #225, opened 2016-04-25
make the java 8 requirement more explicit and replace in README managed-vm with AE Flexible Environement

aozarov:


README.md
```diff
@@ -1,7 +1,7 @@
 appengine-java-vm-runtime
 =========================
 
-Complete code source of the Google App Engine [Java Managed VM](https://cloud.google.com/appengine/docs/managed-vms/) Docker image.
+Complete code source of the Google App Engine [Flexible Environment](https://cloud.google.com/appengine/docs/flexible/) Docker image.
```
chanseokoh: You will notice [here](https://cloud.google.com/appengine/docs/flexible/) that the words "flexible environment" are not capitalized unless they are used in a title. It's a marketing decision (on which I don't have a good understanding.)

aozarov: Done.

PR #227, opened 2016-04-28
Upgrade to the 1.9.36 GAE APIs.

ludoch:


pom.xml
```diff
@@ -126,6 +126,9 @@
           <groupId>com.spotify</groupId>
           <artifactId>docker-maven-plugin</artifactId>
           <version>0.3.2</version>
+          <configuration>
```
aozarov: what is that for?

ludoch: Good catch...
I am using Docker Mac beta early access, and this is a new setting for this
non standard docker.
I need to remove my specific setting until we standardize with this Docker
Mac version...

Issue #230, opened 2016-05-05
Provide alternate request ticket behaviours


gregw:
Currently the async-support branch installs a [HttpOutput.Interceptor](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/async-support/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java#L362-L380)  that clears the request ticket as the response is committed.

This behaviour should be made configurable, as the request ticket may remain valid while the response is being written.

Furthermore, an configuration option could be provided to mimic the buffering behaviour of the master branch by buffer content here and only flushing when the full request lifecycle is complete.

aozarov:
For synchronous calls is it guaranteed that the `Interceptor.write` callback is only going to be called once at the end of the method? Otherwise shouldn't it consider the `complete` param or is that the reason you were suggesting to mimic the buffering behavior of the master branch?

Do we still think that using the request ticket performs better than using the global ticket?
Speaking with the VM people they think they should have the same performance characteristics and if
that is true. /cc @ludoch

gregw:
The interceptor write method will be called every time the application does a flush OR writes enough data to fill up the aggregation buffer inside our HttpOutput class.   Thus the first write is seen as the commit.      A write with complete==true will always be done, even if it is with an empty buffer (used by a close), so that can indicate the end of writing, but as that could be an explicit close done by either a synchronous or async servlet, it is still possible for API calls to be made after the complete==true write.

API calls can theoretically continue to be made for sync calls until the `request destroyed` listener callback or for async calls until the `AsyncContext.onComplete` callback.

Even if performance is the same between global and request ticket, it is desirable if we can use the request ticket so that trace and logging show the APIs correctly associated with a request.

PR #233, opened 2016-05-09
Issue #230

gregw:
Implemented #230
Factored out VmRuntimeInterceptor
Use beta settings to configure to clear the request ticket: never, after commit, after close

@ludoch is configuration via the beta settings in appengine-web.xml a good way to support this?

### DIFF
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeInterceptor.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeInterceptor.java
@@ -1,0 +1,92 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS-IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.apphosting.vmruntime.jetty9;
+
+import com.google.apphosting.utils.config.AppEngineWebXml;
+import com.google.apphosting.vmruntime.VmApiProxyEnvironment;
+
+import org.eclipse.jetty.server.HttpOutput;
+import org.eclipse.jetty.util.Callback;
+
+import java.nio.ByteBuffer;
+
+/**
+ * HttpOutput Interceptor.
+ *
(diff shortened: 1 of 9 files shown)


gregw:
I signed it!

ludoch:
1/ the beta settings were not meant to configure that type of behavior.
2/ based on Arie's finding talking to engineers for the gae new stack, the request ticket will disappear in a very near future, and only the global ticket will be supported.   I am not sure we want to continue trying to optimize for the 2 targets... 
Google hangout soon to chat more?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -361,23 +366,7 @@ public RequestContext getRequestContext(Request baseRequest) {
           VmRuntimeUtils.ONE_DAY_IN_MILLIS, defaultEnvironment);
       final HttpOutput httpOutput = request.getResponse().getHttpOutput();
       final HttpOutput.Interceptor nextOutput = httpOutput.getInterceptor();
-      httpOutput.setInterceptor(new Interceptor() {
-        @Override
-        public void write(ByteBuffer content, boolean complete, Callback callback) {
-          requestSpecificEnvironment.clearTicket(); // Request ticket not valid after commit
```
aozarov: Help me understand. So, in the current code (where we don't buffer/delay the applications writes) this is going to clear the ticket after the first write to the servlet's outputstream?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeInterceptor.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Override
+  public void write(ByteBuffer content, boolean last, Callback callback) {
+    if (clearTicket == ClearTicket.ON_COMMIT)
+      environment.clearTicket();
+
+    next.write(content, last, callback);
+
+    if (clearTicket == ClearTicket.ON_LAST && last)
```
aozarov: for synchronous calls last is set to `true` once the http request completes?

What happens for asynchronous calls, do they use the same set of interceptors and if so, `last` will be set to `false` only once `AsyncContext.complete` is called?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeInterceptor.java
```diff
@@ -0,0 +1,92 @@
...
+
+  public VmRuntimeInterceptor(VmApiProxyEnvironment environment, HttpOutput.Interceptor next) {
+    this.environment = environment;
+    this.next = next;
+    this.clearTicket = dftClearTicket;
+  }
+
+  @Override
+  public void write(ByteBuffer content, boolean last, Callback callback) {
+    if (clearTicket == ClearTicket.ON_COMMIT)
```
aozarov: any reason to clear it before calling next?

gregw:
replaced by #237

Issue #242, opened 2016-05-18
Reduce log level for VmApiProxyDelegate


aozarov:
We should change VmApiProxyDelegate logging (e.g. for `makeApiCall`) to DEBUG (Using `LogLevel.FINE`).  Currently every request would have at least 4 separate log entries for, I guess, the session management (2 for datastore and 2 for memcache).

Also, I think we should start adding `'traceId' to the structured log metadata (which hopefully will allow combining related log lines).

gregw:
See comment on #240 for traceId

PR #244, opened 2016-05-19
Issue #242 Less verbose logging

gregw:
Reduced the verbosity of logs in VmApiProxyDelegate for #242
Also reviewed log levels in AppEngineAuthentication (and fixed some javadoc errors).

### DIFF
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
@@ -161,11 +161,13 @@ public class VmApiProxyDelegate implements ApiProxy.Delegate<VmApiProxyEnvironme
     try {
       byte responseData[] = runSyncCall(environment, packageName, methodName, requestData, timeoutMs);
       long end = System.currentTimeMillis();
-      logger.log(Level.FINE, String.format(
+      if (logger.isLoggable(Level.FINE)) {
+        logger.log(Level.FINE, String.format(
           "Service bridge API call to package: %s, call: %s, of size: %s " +
-          "complete. Service bridge status code: %s; response " +
-          "content-length: %s. Took %s ms.", packageName, methodName, requestData.length, 200,
+            "complete. Service bridge status code: %s; response " +
+            "content-length: %s. Took %s ms.", packageName, methodName, requestData.length, 200,
           responseData.length, (end - start)));
+      }
       
       // TODO Remove HACK TO FIX USER_SERVICE ISSUE #164
       // Disable with -DUserServiceLocalSchemeHost=false
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
@@ -161,7 +161,7 @@ public class VmApiProxyDelegate implements ApiProxy.Delegate<VmApiProxyEnvironme
     try {
       byte responseData[] = runSyncCall(environment, packageName, methodName, requestData, timeoutMs);
       long end = System.currentTimeMillis();
-      logger.log(Level.INFO, String.format(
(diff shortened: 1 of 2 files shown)


appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
```diff
@@ -161,7 +161,7 @@ public VmApiProxyDelegate() {
     try {
       byte responseData[] = runSyncCall(environment, packageName, methodName, requestData, timeoutMs);
       long end = System.currentTimeMillis();
-      logger.log(Level.INFO, String.format(
+      logger.log(Level.FINE, String.format(
```
aozarov: Do you think it is worth a `isLoggable` check?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppEngineAuthentication.java
```diff
@@ -354,15 +354,14 @@ public void setIdentityService(IdentityService identityService) {
     }
 
     /**
-     * Validate a user identity. Validate that a UserIdentity previously created by a call to
-     * {@link #login(String, Object)} is still valid.
+     * Validate a user identity. Validate that a UserIdentity is still valid.
```
aozarov: 1) I think "Validate that a UserIdentity is still valid." should have been enough.
2) I think such doc change should have been applied on `LoginService` (as it does not seem to be specific to this impl).
3) I would normally leave out JavaDoc for methods that I override/implement unless they provide specifics for the implementation. In this case we just throw `UnsupportedException`

aozarov:
Looks good. 2 nits.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
gregw: indeed, will do

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppEngineAuthentication.java
gregw: all good points.  I was just doing minimal edits to remove the hard javadoc errors of broken links.

gregw:
denitted

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
```diff
@@ -161,11 +161,13 @@ public VmApiProxyDelegate() {
     try {
       byte responseData[] = runSyncCall(environment, packageName, methodName, requestData, timeoutMs);
       long end = System.currentTimeMillis();
-      logger.log(Level.INFO, String.format(
+      if (logger.isLoggable(Level.FINE)) {
+        logger.log(Level.FINE, String.format(
           "Service bridge API call to package: %s, call: %s, of size: %s " +
-          "complete. Service bridge status code: %s; response " +
-          "content-length: %s. Took %s ms.", packageName, methodName, requestData.length, 200,
+            "complete. Service bridge status code: %s; response " +
```
aozarov: No need to change as it is consistent with the file. So comment is more of an FYI about the line wrapping which should come before the '+' (https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break). I am surprised that the formatter didn't fix that for you.

aozarov:
BTW, any preference of CI Travis vs CircleCi ?

gregw:
@aozarov I'll defer to @joakime for a more definitive answer.     We've mostly used Jenkins and it is OK, but slightly confusing UI and I've not yet seen it integrated with github.    

I have used circle a little bit and have mixed feelings about it - it can be nice and simple, but if it does not work then it can be hard to configure.   It can also have problems with large builds and reviewing the output of them can lock up your browser.

No recent experience with travis.

PR #250, opened 2016-05-24
Applying a Google formatter

aozarov:
I used: `find . - name '*.java' | xargs google-java-format -i`
Also, I fixed missing braces (as formatter didn't do that - see google/google-java-format#51)

I manually sampled many of the changes. Most looks good. There are few cases, where I think
the formatter can do a better job but that could be fixed later.


ludoch:
Scrolled all the way.....
LGTM

PR #254, opened 2016-05-24
sort imports acording to the Google style

aozarov:


ludoch:
Which tool did you use?

gregw:
@aozarov I think it is a good idea to get all the code reformatted like this.  However once done, we need very good style configurations for eclipse and IDEA, so that reformats will produce no changes at all.

aozarov:
@ludoch , I used the internal sort tool
```
find . -name '*.java' | xargs sort_java_imports.py 
```

@gregw, indeed. see [google-java-format](https://github.com/google/google-java-format)
This is the rule for sort order: https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing
Both eclipse and intelliJ have options to specify this order (I can share my settings with you if you want). Also, they have an option to cleanup unused import statements.

aozarov:
Also, looks like sort order would be part of the google-java-format once this [bug](https://github.com/google/google-java-format/issues/42) is fixed.

ludoch:
lgtm

Issue #229, opened 2016-05-05
Are session loaded lazily (upon the first call of getSession) and only saved if dirty?


aozarov:
/cc @gregw

janbartel:
When a request enters the handler chain, the jetty SessionHandler will find a session cookie or jsession id encoded url on the request and ask a SessionManager to look up the session. In the appengine compat code the Session Manager lookup goes to memcache first and then falls through to gcloud datastore.  The data of the session is populated into a fresh AppEngineSession object.  

A session is written to gcloud datastore and memcache when the request is exiting the server, as it is about to be destroyed. It is only written out iff the session is dirty, which occurs if either:

- set/removeAttribute was called on the session

- time remaining until the session expires is at least 75% of the session timeout

The latter has the effect of avoiding frequent writes of a session where all that changed was the expiry time (last access time).

Because session objects are not shared amongst simultaneous requests with the same session id, it is non-deterministic what is stored as the requests exit: if request A sets attribute X and request B sets attribute Y, it depends on what order the requests complete as to whether the stored session will contain attribute X or Y (it is impossible for it to contain both X and Y).

aozarov:
I see, thanks. So, it looks like if I want to have a test that shows the difference between session on or off I actually need to use the session.

BTW, what will happen if I use the session but `appengine-web.xml` has `<sessions-enabled>false</sessions-enabled>`?

janbartel:
More details on sessions are in my posts on issue #228.  To recap part of what I said there, if <sessions-enabled>false</sessions-enabled> in appengine-web.xml, then jetty falls back to using our in-memory-only, non-clustered HashSessionManager.

aozarov:
Yes, thank!

I do wonder however if that is not confusing. In that case I would have called the option persistent-sessions-enabled instead (as to me sessions-enabled false suggest no sessions, hence session cookie  never sent back).

gregw:
@aozarov  The fallback to HashSessionManager predates our involvement, so I'm not entirely sure why it was done (and I agree it is confusing).

it would be easy to totally remove the SessionHandler (or not add it in the first place) if sessions-enabled is false.      This would mean a session cookie would never be set and getSession would always return null.

If that is desirable, label this issue as an enhancement and assign to me and I'll make it so!

janbartel:
[...]

PR #255, opened 2016-05-25
Async support 229

janbartel:
A very minimal session manager for when sessions-enabled=false as per enhancement request #229.

### DIFF
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
@@ -54,7 +54,9 @@ public class NoOpSession implements HttpSession {
    * Finish a request
    */
   protected void complete() {
-    attributes.clear();
+    synchronized (this) {
+      attributes.clear();
+    }
   }
 
   @Override
@@ -119,7 +121,7 @@ public class NoOpSession implements HttpSession {
   public Enumeration<String> getAttributeNames() {
     synchronized (this) {
       checkValid();
-      List<String> names = new ArrayList<String>(attributes.keySet());
+      List<String> names = new ArrayList<>(attributes.keySet());
       return Collections.enumeration(names);
     }
   }
@@ -129,7 +131,7 @@ public class NoOpSession implements HttpSession {
   public String[] getValueNames() {
     synchronized (this) {
       checkValid();
-      List<String> names = new ArrayList<String>(attributes.keySet());
+      List<String> names = new ArrayList<>(attributes.keySet());
       if (names.isEmpty()) {
         return new String[0];
       }
@@ -140,6 +142,7 @@ public class NoOpSession implements HttpSession {
   @Override
   public void setAttribute(String name, Object value) {
     synchronized (this) {
+      checkValid();
       attributes.put(name, value);
     }
   }
@@ -153,6 +156,7 @@ public class NoOpSession implements HttpSession {
   @Override
   public void removeAttribute(String name) {
     synchronized (this) {
+      checkValid();
       attributes.remove(name);
     }
   }
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
@@ -27,8 +27,6 @@ import javax.servlet.http.HttpSessionContext;
 
 
 /**
- * NoOpSession
- *
  * A minimal implementation of HttpSession, just sufficient to keep and id attributes within a
  * single request.
  */
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
@@ -27,7 +27,6 @@ import javax.servlet.ServletContext;
 import javax.servlet.http.HttpSession;
 import javax.servlet.http.HttpSessionContext;
 
-
 /**
  * A minimal implementation of HttpSession, just sufficient to use attributes within a
  * single request.
@@ -57,7 +56,6 @@ public class NoOpSession implements HttpSession {
   protected void complete() {
     attributes.clear();
   }
-
 
   @Override
   public long getCreationTime() {
@@ -89,8 +87,7 @@ public class NoOpSession implements HttpSession {
   }
 
   @Override
-  public void setMaxInactiveInterval(int interval) {
-  }
+  public void setMaxInactiveInterval(int interval) {}
 
   @Override
   public int getMaxInactiveInterval() {
@@ -187,9 +184,8 @@ public class NoOpSession implements HttpSession {
     checkValid();
     return true;
   }
-  
-  private void checkValid () throws IllegalStateException {
-    if (!isValid())
-      throw new IllegalStateException("Session "+clusterId+" invalid");
+
+  private void checkValid() throws IllegalStateException {
+    if (!isValid()) throw new IllegalStateException("Session " + clusterId + " invalid");
   }
 }
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
@@ -1,0 +1,187 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS-IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+
+package com.google.apphosting.runtime.jetty9;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpSession;
+import javax.servlet.http.HttpSessionContext;
+
+
+/**
+ * NoOpSession
+ *
+ * A minimal implementation of HttpSession, just sufficient to keep and id
+ * attributes within a single request.
+ * 
+ */
+public class NoOpSession implements HttpSession {
+
+  protected ServletContext _context;
+  protected String _clusterId;
+  protected String _nodeId;
+  protected boolean _valid = true;
+  private final Map<String,Object> _attributes=new HashMap<String, Object>();
+
+
+  /**
+   * @param id
+   * @param nodeId
+   * @param context
+   */
+  protected NoOpSession(String id, String nodeId, ServletContext context) {
+      _clusterId = id;
+      _nodeId = nodeId;
+      _context = context;
+  }
+
+
+  /**
+   * Finish a request
+   */
+  public void complete () {
+      _attributes.clear();
+  }
+
+
+  /** 
+   * @see javax.servlet.http.HttpSession#getCreationTime()
+   */
+  @Override
+  public long getCreationTime() {
+      return 0;
+  }
+
+
+  @Override
+  public String getId() {
+      return _clusterId;
+  }
+
+  /**
+   * @return the session id with worker suffix
+   */
+  public String getNodeId() {
+      return _nodeId;
+  }
+
+
+  @Override
+  public long getLastAccessedTime() {
+    return 0;
+  }
+
+
+  @Override
+  public ServletContext getServletContext() {
+    return _context;
+  }
+
+
+  @Override
+  public void setMaxInactiveInterval(int interval) {
+  }
+
+
+  @Override
+  public int getMaxInactiveInterval() {
+    return 0;
+  }
+
+
+  @Override
+  public HttpSessionContext getSessionContext() {
+    return null;
+  }
+
+
+  @Override
+  public Object getAttribute(String name) {
+    return _attributes.get(name);
+  }
+
+
+  @Override
+  public Object getValue(String name) {
+    return getAttribute(name);
+  }
+
+
+  @Override
+  public Enumeration<String> getAttributeNames() {
+    List<String> names=_attributes==null?Collections.EMPTY_LIST:new ArrayList<String>(_attributes.keySet());
+    return Collections.enumeration(names);
+  }
+
+
+  @Override
+  public String[] getValueNames() {
+    List<String> names=_attributes==null?Collections.EMPTY_LIST:new ArrayList<String>(_attributes.keySet());
+    if (names.isEmpty())
+      return null;
+    return names.toArray(new String[names.size()]);
+  }
+
+
+  @Override
+  public void setAttribute(String name, Object value) {
+   _attributes.put(name, value);
+  }
+
+
+  @Override
+  public void putValue(String name, Object value) {
+    setAttribute(name,value);
+  }
+
+
+  @Override
+  public void removeAttribute(String name) {
+   _attributes.remove(name);
+  }
+
+
+  @Override
+  public void removeValue(String name) {
+    removeAttribute(name);
+  }
+
+
+  @Override
+  public void invalidate() {
+      _valid = false;
+      _attributes.clear();
+  }
+
+  /**
+   * @return true if the session has not been invalidated
+   */
+  public boolean isValid () {
+      return _valid;
+  }
+
+  @Override
+  public boolean isNew() {
+    return true;
+  }
+
+}
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
@@ -186,6 +186,8 @@ public class NoOpSession implements HttpSession {
   }
 
   private void checkValid() throws IllegalStateException {
-    if (!isValid()) throw new IllegalStateException("Session " + clusterId + " invalid");
+    if (!isValid()) {
+      throw new IllegalStateException("Session " + clusterId + " invalid");
+    }
   }
 }
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
@@ -12,7 +12,6 @@
  * the License.
  */
 
-
 package com.google.apphosting.runtime.jetty9;
 
 import java.util.ArrayList;
@@ -30,9 +29,8 @@ import javax.servlet.http.HttpSessionContext;
 /**
  * NoOpSession
  *
- * A minimal implementation of HttpSession, just sufficient to keep and id
- * attributes within a single request.
- * 
+ * A minimal implementation of HttpSession, just sufficient to keep and id attributes within a
+ * single request.
  */
 public class NoOpSession implements HttpSession {
 
@@ -40,8 +38,7 @@ public class NoOpSession implements HttpSession {
   protected String _clusterId;
   protected String _nodeId;
   protected boolean _valid = true;
-  private final Map<String,Object> _attributes=new HashMap<String, Object>();
-
+  private final Map<String, Object> _attributes = new HashMap<String, Object>();
 
   /**
    * @param id
@@ -49,39 +46,38 @@ public class NoOpSession implements HttpSession {
    * @param context
    */
   protected NoOpSession(String id, String nodeId, ServletContext context) {
-      _clusterId = id;
-      _nodeId = nodeId;
-      _context = context;
+    _clusterId = id;
+    _nodeId = nodeId;
+    _context = context;
   }
-
 
   /**
    * Finish a request
    */
-  public void complete () {
-      _attributes.clear();
+  public void complete() {
+    _attributes.clear();
   }
 
 
-  /** 
+  /**
    * @see javax.servlet.http.HttpSession#getCreationTime()
    */
   @Override
   public long getCreationTime() {
-      return 0;
+    return 0;
   }
 
 
   @Override
   public String getId() {
-      return _clusterId;
+    return _clusterId;
   }
 
   /**
    * @return the session id with worker suffix
    */
   public String getNodeId() {
-      return _nodeId;
+    return _nodeId;
   }
 
 
@@ -128,35 +124,40 @@ public class NoOpSession implements HttpSession {
 
   @Override
   public Enumeration<String> getAttributeNames() {
-    List<String> names=_attributes==null?Collections.EMPTY_LIST:new ArrayList<String>(_attributes.keySet());
+    List<String> names = _attributes == null
+      ? Collections.EMPTY_LIST
+      : new ArrayList<String>(_attributes.keySet());
     return Collections.enumeration(names);
   }
 
 
   @Override
   public String[] getValueNames() {
-    List<String> names=_attributes==null?Collections.EMPTY_LIST:new ArrayList<String>(_attributes.keySet());
-    if (names.isEmpty())
+    List<String> names = _attributes == null
+      ? Collections.EMPTY_LIST
+      : new ArrayList<String>(_attributes.keySet());
+    if (names.isEmpty()) {
       return null;
+    }
     return names.toArray(new String[names.size()]);
   }
 
 
   @Override
   public void setAttribute(String name, Object value) {
-   _attributes.put(name, value);
+    _attributes.put(name, value);
   }
 
 
   @Override
   public void putValue(String name, Object value) {
-    setAttribute(name,value);
+    setAttribute(name, value);
   }
 
 
   @Override
   public void removeAttribute(String name) {
-   _attributes.remove(name);
+    _attributes.remove(name);
   }
 
 
@@ -168,15 +169,15 @@ public class NoOpSession implements HttpSession {
 
   @Override
   public void invalidate() {
-      _valid = false;
-      _attributes.clear();
+    _valid = false;
+    _attributes.clear();
   }
 
   /**
    * @return true if the session has not been invalidated
    */
-  public boolean isValid () {
-      return _valid;
+  public boolean isValid() {
+    return _valid;
   }
 
   @Override
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
@@ -1,15 +1,17 @@
-/**
+/*
  * Copyright 2016 Google Inc. All Rights Reserved.
  *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
- * in compliance with the License. You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *       http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software distributed under the License
- * is distributed on an "AS-IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied. See the License for the specific language governing permissions and limitations under
- * the License.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 
 package com.google.apphosting.runtime.jetty9;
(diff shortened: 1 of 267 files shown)


gregw:
I signed it!

gregw:
I signed it!

aozarov:
I think @janbartel needs to sign as well (probably using janb@webtide.com).
Also, I don't understand why you need to keep replying "I signed it!" I assume all your commits are using "Greg Wilkins <gregw@webtide.com>" and that this name and email are used for singing agreement.
Maybe @ludoch knows or sending an email to github-admin@google.com may help.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.List;
+import java.util.Map;
+
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpSession;
+import javax.servlet.http.HttpSessionContext;
+
+
+/**
+ * NoOpSession
```
aozarov: remove and move the lines bellow to be the class level Javadoc description.

janbartel:
@aozarov I signed the CLA months ago. Not sure why it keeps complaining.

janbartel:
I signed it!

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
aozarov: Missing the 5 spaces.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.List;
+import java.util.Map;
+
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpSession;
+import javax.servlet.http.HttpSessionContext;
+
+
+/**
+ * A minimal implementation of HttpSession, just sufficient to keep and id attributes within a
```
aozarov: I find "keep and id attributes" to be a bit cryptic. Maybe "use"?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+ * single request.
+ */
+public class NoOpSession implements HttpSession {
+
+  protected ServletContext _context;
```
aozarov: Don't use the _ prefix. see: https://google.github.io/styleguide/javaguide.html#s5.2.5-non-constant-field-names and https://google.github.io/styleguide/javaguide.html#s5-naming

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+  protected String _clusterId;
+  protected String _nodeId;
+  protected boolean _valid = true;
+  private final Map<String, Object> _attributes = new HashMap<String, Object>();
```
aozarov: You can use the diamond here.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
```
aozarov: either fill the javadoc or remove the empty template.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+  protected NoOpSession(String id, String nodeId, ServletContext context) {
+    _clusterId = id;
+    _nodeId = nodeId;
+    _context = context;
+  }
+
+  /**
+   * Finish a request
+   */
+  public void complete() {
```
aozarov: private?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+    _attributes.clear();
+  }
+
```
aozarov: Only one line separator (here and bellow).

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+   * @see javax.servlet.http.HttpSession#getCreationTime()
```
aozarov: I am not sure this (auto-generated javadoc?) is useful (I am sure it is not needed by the Javadoc generator). Also, it is not consistent with the other methods.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+
+  @Override
+  public String getId() {
+    return _clusterId;
+  }
+
+  /**
+   * @return the session id with worker suffix
+   */
+  public String getNodeId() {
```
aozarov: is it planned to be called from outside of this class?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+    if (names.isEmpty()) {
+      return null;
+    }
+    return names.toArray(new String[names.size()]);
+  }
+
+
+  @Override
+  public void setAttribute(String name, Object value) {
+    _attributes.put(name, value);
```
aozarov: is access guaranteed to be synchronized?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+
+
+  @Override
+  public long getLastAccessedTime() {
+    return 0;
+  }
+
+
+  @Override
+  public ServletContext getServletContext() {
```
aozarov: do you need `@SuppressWarnings("deprecation")` here and for `xxxValue` methods or is the Java 8 compiler smart enough not to warn about those?

aozarov: Actually I think if the compiler does warn about it we should probably add `@Deprecated` instead.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+
+  @Override
+  public String[] getValueNames() {
+    List<String> names = _attributes == null
+      ? Collections.EMPTY_LIST
+      : new ArrayList<String>(_attributes.keySet());
+    if (names.isEmpty()) {
+      return null;
```
aozarov: is it expected to return `null` in this case or an empty array?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+
+  @Override
+  public void removeValue(String name) {
+    removeAttribute(name);
+  }
+
+
+  @Override
+  public void invalidate() {
+    _valid = false;
```
aozarov: should all methods, including this one, have a `Preconditions.checkState(_valid, "Session is  not valid");` ?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+   */
+  public boolean isValid() {
+    return _valid;
+  }
+
+  @Override
+  public boolean isNew() {
+    return true;
+  }
+
```
aozarov: remove empty line here.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionIdManager.java
```diff
@@ -0,0 +1,64 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
aozarov: Fix copyright.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionIdManager.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.eclipse.jetty.server.session.AbstractSessionIdManager;
+
+/**
+ * Minimal implementation of the SessionIdManager, just enough to generate new session ids.
+ */
+public class NoOpSessionIdManager extends AbstractSessionIdManager {
+
+  /**
+   * @see org.eclipse.jetty.server.SessionIdManager#idInUse(java.lang.String)
```
aozarov: here and bellow. this is unnecessary/redundant as the generated Javadoc will include this information.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionIdManager.java
```diff
@@ -0,0 +1,64 @@
...
+  public void invalidateAll(String id) {
+  }
+
+  /**
+   * @see org.eclipse.jetty.server.session.AbstractSessionIdManager#renewSessionId(java.lang.String,
+   * java.lang.String, javax.servlet.http.HttpServletRequest)
+   */
+  @Override
+  public void renewSessionId(String oldClusterId, String oldNodeId, HttpServletRequest request) {
+    return; //don't support this
```
aozarov: what is the point of this `return;`?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
```
aozarov: fix copyright.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+  @Override
+  public Object getValue(String name) {
+    return getAttribute(name);
+  }
+
+
+  @Override
+  public Enumeration<String> getAttributeNames() {
+    List<String> names = _attributes == null
```
aozarov: here and bellow:
* I don't think `_attributes` can ever be `null`
* You can use `Collections.emptyList()`
* I think you can use the diamond here as well.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.SessionIdManager;
+import org.eclipse.jetty.server.SessionManager;
+import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.server.session.SessionHandler;
+import org.eclipse.jetty.util.component.ContainerLifeCycle;
+
+/**
+ * Does the bare minimum to create a NoOpSession. This session manager should be used only when
+ * app-cfg.xml has disabled sessions. The purpose of it is to provide an implementation with just
```
aozarov: `app-cfg.xml` ?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+ * enough compliance with the servlet session api so that frameworks that require the session api
+ * will still work, albeit without the costs associated with distributed persistent sessions (ie
+ * sessions enabled in app-cfg.xml).
+ */
+public class NoOpSessionManager extends ContainerLifeCycle implements SessionManager {
+
+  private SessionIdManager _idManager;
```
aozarov: avoid the _ prefix.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+  private SessionHandler _handler;
+  private ServletContext _context;
+
+  /**
```
aozarov: javadoc not useful. Actually, the constructor itself is also not needed.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+        addBean(_idManager, false);
+      }
+    }
+    super.doStart();
+  }
+
+
+  @Override
+  public HttpSession getHttpSession(String id) {
+    return null;
```
aozarov: Can this ever be called within the same request or when forwarding a request?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  @Override
+  public HttpSession newHttpSession(HttpServletRequest request) {
+
```
aozarov: remove empty line

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+    _handler = handler;
+  }
+
+  @Override
+  public void addEventListener(EventListener listener) {
+  }
+
+  @Override
+  public void removeEventListener(EventListener listener) {
+
```
aozarov: remove empty line

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  @Override
+  public void clearEventListeners() {
+  }
+
+  @Override
+  public HttpCookie getSessionCookie(HttpSession session, String contextPath,
```
aozarov: the preferred wrap in this case would be:
```
public HttpCookie getSessionCookie(
    HttpSession session, String contextPath, boolean requestIsSecure) {
```

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+   *
+   */
+  public NoOpSessionManager() {
+  }
+
+
+  @Override
+  protected void doStart() throws Exception {
+    _context = ContextHandler.getCurrentContext();
+    final Server server = _handler.getServer();
```
aozarov: We avoid using final for local variables or params.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  @Override
+  public boolean isUsingURLs() {
+    return false;
+  }
+
+  @Override
+  public Set<SessionTrackingMode> getDefaultSessionTrackingModes() {
+    return null;
```
aozarov: Not much documentation [here](https://github.com/eclipse/jetty.project/blob/jetty-9.3.x/jetty-server/src/main/java/org/eclipse/jetty/server/SessionManager.java#L290) but I assume `null` is an expected return value?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+  public boolean isCheckingRemoteSessionIdEncoding() {
+    return true;
+  }
+
+  @Override
+  public void setCheckingRemoteSessionIdEncoding(boolean remote) {
+  }
+
+  @Override
+  public void renewSessionId(String oldClusterId, String oldNodeId, String newClusterId,
```
aozarov: all params should be wrapped in one line. If that lines does not fit wrapping should be each param in its own line +4 spaces. the google-java-formatter should do that.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
```diff
@@ -0,0 +1,237 @@
...
+                             String newNodeId) {
+  }
+
```
aozarov: remove empty line.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
```diff
@@ -46,10 +46,9 @@
 import com.google.apphosting.runtime.SessionStore;
 
 /**
- * Implements the Jetty {@link AbstractSessionManager} and, as an inner class,
- * {@link HashSessionIdManager} for our context. The session manager has to check the provided
- * {@link SessionStore SessionStores} to find sessions.
- *
+ * Implements the Jetty {@link AbstractSessionManager} and, as an inner class, {@link
```
aozarov: I think avoiding the line wrap inside a tag which was done before is nicer.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
```diff
@@ -107,11 +106,9 @@ public String generateNewId() {
 
   /**
    * A session implementation using the provided list of {@link SessionStore SessionStores} to store
-   * attributes. Expiration is also stored.
-   * <p>
-   * An instance of this class may be used simultaneously be multiple request threads. We use
-   * synchronization to guard access to sessionData and the parent object state.
-   *
+   * attributes. Expiration is also stored. <p> An instance of this class may be used simultaneously
```
aozarov: new line before `<p>`

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
```diff
@@ -156,11 +153,11 @@ public void renewId(HttpServletRequest request) {
 
       // generate a new id
       String newClusterId =
-          ((SessionIdManager) getSessionManager().getSessionIdManager())
-              .newSessionId(request.hashCode());
+        ((SessionIdManager) getSessionManager().getSessionIdManager())
```
aozarov: wrapping is +4 not +2 spaces (before was good).

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
```diff
@@ -220,10 +217,10 @@ protected void save(boolean force) {
             delay *= 2;
           }
           logger.log(
-              Level.SEVERE, String.format("Session %s not saved: too many attempts", getId()));
+            Level.SEVERE, String.format("Session %s not saved: too many attempts", getId()));
```
aozarov: same like above.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
```diff
@@ -220,10 +217,10 @@ protected void save(boolean force) {
...
         } catch (DeadlineExceededException e) {
           logger.log(
-              Level.SEVERE, String.format("Session %s not saved: too many timeouts", getId()), e);
+            Level.SEVERE, String.format("Session %s not saved: too many timeouts", getId()), e);
```
aozarov: the same. Was there anything changed in this file other than formatting?
If not, I would revert it as a side of the removal of the empty `@param` at the bottom
of the file, most seems to violate the guideline.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -341,15 +342,19 @@ public void init(String appengineWebXmlFile) throws AppEngineConfigException, IO
       // No need to configure the session manager.
       return;
     }
-    AbstractSessionManager sessionManager;
+    org.eclipse.jetty.server.SessionManager sessionManager;
```
aozarov: does`AbstractSessionManager` import statement needed anymore?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,186 @@
...
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpSession;
+import javax.servlet.http.HttpSessionContext;
+
+
+/**
+ * A minimal implementation of HttpSession, just sufficient to keep and id attributes within a
+ * single request.
+ */
+public class NoOpSession implements HttpSession {
```
aozarov: A unit tests would be nice. Even though this class does not do much it would be good for making sure that background compatibility is not changed without a test failing.

aozarov:
Done with my review. Lots of nit picks and few questions.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
janbartel: The javadoc for the servlet api is silent on this point. See http://docs.oracle.com/javaee/7/api/javax/servlet/ServletContext.html#getEffectiveSessionTrackingModes--

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
janbartel: This is called only when a request enters a new context. Thus if a forward is done to the same context, it isn't called. AFAIK forwards to a different context aren't supported in compat because there is only ever 1 context deployed.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSessionManager.java
aozarov: I see, though not explicit when I read "set of the session tracking modes supported by default for this ServletContext" I would expect empty set and not null if nothing is supported. Fine as is though.

gregw:
@aozarov Both Jan an I have signed the CLA and we both only have a single github identity.    The CLA-bot sometimes sees us as having signed, but not at other times.  We can't spot the pattern, but assume it is a bug in the bot?

aozarov:
@gregw I am in content with the CLA people and they asked me to ask you that: "what is the email for the google group was that was created for this corporate CLA?"

aozarov:
@janbartel there are some conflicts that needs to be fixed.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  @Override
+  public boolean isNew() {
+    checkValid();
+    return true;
+  }
+
+  private void checkValid() throws IllegalStateException {
+    if (!isValid()) throw new IllegalStateException("Session " + clusterId + " invalid");
```
aozarov: Any reason for not using Guava's Preconditions? could be one-liner.
If you prefer as is then change it to:
```java
if (!isValid()) {
  throw new IllegalStateException("Session " + clusterId + " invalid")
}
```

otherwise you could do:
```java
Preconditions.checkState(isValid(), "Session %s invalid", clusterId);
```

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,191 @@
...
+  @Override
+  public Object getValue(String name) {
+    return getAttribute(name);
+  }
+
+  @Override
+  public Enumeration<String> getAttributeNames() {
+    synchronized (this) {
+      checkValid();
+      List<String> names = new ArrayList<String>(attributes.keySet());
```
aozarov: here and bellow you can use `<>`

gregw:
@aozarov The google group is gae-contributors@webtide.com

gregw:
@aozarov This is still failing the build because of the checkformat step.   However the classes have been formatted with google-java-format-1.0-all-deps.jar  So there is a difference between the auto formatting and the checker!   Do you want us to manually fix?

aozarov:
The failure I see are related to missing brackets and sort import. As mention before the tool does not yet handle them (internal or externally), so those should be manually fixed (and the tool is not suppose to revert it). Are there any more?

gregw:
@aozarov I fixed most of the formatting errors, but IP validation is still broken in the PR

aozarov:
What do you mean by "IP validation"?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,193 @@
...
+
+  @Override
+  public long getCreationTime() {
+    checkValid();
+    return 0;
+  }
+
+  @Override
+  public String getId() {
+    return clusterId;
```
aozarov: For being "compatible" aren't you expect to call `chackValid` in all the method you override that suggests it in the javadoc (seems that there are few methods that do not specify it such as `getServletContext`).

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,193 @@
...
+    this.clusterId = id;
+    this.nodeId = nodeId;
+    this.context = context;
+  }
+
+  /**
+   * Finish a request
+   */
+  protected void complete() {
+    attributes.clear();
```
aozarov: is it safe to call it without synchronized?

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
```diff
@@ -0,0 +1,193 @@
...
+      if (names.isEmpty()) {
+        return new String[0];
+      }
+      return names.toArray(new String[names.size()]);
+    }
+  }
+
+  @Override
+  public void setAttribute(String name, Object value) {
+    synchronized (this) {
```
aozarov: checkValid (also see my previous comment about it).

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
```diff
@@ -148,8 +148,7 @@ public boolean isDirty() {
 
     @Override
     public void renewId(HttpServletRequest request) {
-
-      String oldId = getClusterId();
+      final String oldId = getClusterId();
```
aozarov: remove `final`

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
```diff
@@ -362,9 +362,10 @@ static String lastId() {
    * Constructs a SessionManager
    *
    * @param sessionStoresInWriteOrder The SessionStores in the order to which they should be
-   *        written. When attempting to load a session, the read order is the opposite of the write
-   *        order, so if the write order is A, B, C, when reading we will first consult C, and if
-   *        the session is not found move on to B, and if not found then on to A.
+   *                                  written. When attempting to load a session, the read order is
```
aozarov: should be +4 is the google-java-formatter doing that?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -355,6 +360,11 @@ public boolean isTrustedRemoteAddr(String remoteAddr) {
     return VmRequestUtils.isTrustedRemoteAddr(isDevMode, remoteAddr);
   }
 
+  /**
+   * Get or create the RequestContext for a request.
+   * @param baseRequest The request to scope the context.
```
aozarov: line separator after description (and before tags).

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
gregw: The final was suggested by the checkstyle warning because the assignment is too far from the usage.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/SessionManager.java
gregw: That change wasn't in my IDE history nor was it repeated either by an IDE format or a googlestyle format, at least not from any starting point I tried.  Mystery?   Fixing now.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
janbartel: Looks like I missed checkValid() on set/removeAttribute(), so I'll add those in. Otherwise I've doublechecked with the servlet spec api javadoc and verified that we are doing checkValid() on the correct methods.

aozarov:
Please ping when all comments were addressed.

I think that instead of `final` we can move `String oldId = getClusterId();` to just before we are setting the new cluster (maybe then checkstyle would not complain, even then I believe this is not an enforced check).

Also, I think [this](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/pull/255#discussion-diff-65288249) comment was not addressed.

jetty9-compat-base/src/main/java/com/google/apphosting/runtime/jetty9/NoOpSession.java
janbartel: Protected means it is accessible in the packge in which it is declared, and outside the package only by implementations of the class. So I'm happy with that protection level. What is your suggestion here?

aozarov: If this class is designed to be sub-classed than `protected` is fine. Otherwise, I would
have it package scope if used by unit-test or private if it doesn't. right now it is `public` and I am not sure why.

PR #257, opened 2016-05-25
suppress expected stack traces during unit testing

gregw:
Configure the logger during the unit tests so that unless debug has been requested, the expected stack traces are suppressed.

### DIFF
--- a/appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
+++ b/appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
@@ -42,6 +42,7 @@ import org.apache.http.protocol.HttpContext;
 import org.mockito.Mockito;
 
 import java.io.ByteArrayInputStream;
+import java.io.Closeable;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.util.Arrays;
@@ -196,37 +197,30 @@ public class VmApiProxyDelegateTest extends TestCase {
     byte[] result = null;
     final Double timeoutInSeconds = 10.0;
 
-    Logger logger = Logger.getLogger(VmApiProxyDelegate.class.getName());
-    Level level = logger.getLevel();
-    if (!logger.isLoggable(Level.FINE)) {
-      logger.setLevel(Level.SEVERE);
-    }
-    if (sync) {
-      try {
-        environment.getAttributes().put(VmApiProxyDelegate.API_DEADLINE_KEY, timeoutInSeconds);
-        result =
-            delegate.makeSyncCall(environment, TEST_PACKAGE_NAME, TEST_METHOD_NAME, requestData);
-        fail();
-      } catch (ApiProxyException exception) {
-        assertEquals(expectedException.getClass(), exception.getClass());
-      } finally {
-        logger.setLevel(level);
-      }
-    } else {
-      try {
-        ApiConfig apiConfig = new ApiConfig();
-        apiConfig.setDeadlineInSeconds(timeoutInSeconds);
-        result =
-            delegate
-                .makeAsyncCall(
-                    environment, TEST_PACKAGE_NAME, TEST_METHOD_NAME, requestData, apiConfig)
-                .get();
-        fail();
-      } catch (ExecutionException exception) {
-        // ExecutionException is expected, and make sure the cause is expected as well.
-        assertEquals(expectedException.getClass(), exception.getCause().getClass());
-      } finally {
-        logger.setLevel(level);
+    try (SuppressedLogging l=new SuppressedLogging()) {
+      if (sync) {
+        try {
+          environment.getAttributes().put(VmApiProxyDelegate.API_DEADLINE_KEY, timeoutInSeconds);
+          result =
+              delegate.makeSyncCall(environment, TEST_PACKAGE_NAME, TEST_METHOD_NAME, requestData);
+          fail();
+        } catch (ApiProxyException exception) {
+          assertEquals(expectedException.getClass(), exception.getClass());
+        }
+      } else {
+        try {
+          ApiConfig apiConfig = new ApiConfig();
+          apiConfig.setDeadlineInSeconds(timeoutInSeconds);
+          result =
+              delegate
+              .makeAsyncCall(
+                  environment, TEST_PACKAGE_NAME, TEST_METHOD_NAME, requestData, apiConfig)
+              .get();
+          fail();
+        } catch (ExecutionException exception) {
+          // ExecutionException is expected, and make sure the cause is expected as well.
+          assertEquals(expectedException.getClass(), exception.getCause().getClass());
+        }
       }
     }
     assertNull(result);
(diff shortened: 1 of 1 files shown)


gregw:
I signed it!

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
```diff
@@ -194,6 +196,11 @@ private void callDelegateWithOneError(
     byte[] result = null;
     final Double timeoutInSeconds = 10.0;
 
+    Logger logger = Logger.getLogger(VmApiProxyDelegate.class.getName());
+    Level level = logger.getLevel();
+    if (!logger.isLoggable(Level.FINE)) {
+      logger.setLevel(Level.SEVERE);
```
aozarov: I am not sure why we jump from `FINE` to `SEVERE`. If we want to suppress anything bellow `SEVERE` we should check for `isLoggable(Leve.WARNING)` or if we only care to suppress `FINE` and bellow we should set it to `CONFIG`.

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
```diff
@@ -202,6 +209,8 @@ private void callDelegateWithOneError(
         fail();
       } catch (ApiProxyException exception) {
         assertEquals(expectedException.getClass(), exception.getClass());
+      } finally {
```
aozarov: I would prefer if we had only one finally to handle the logger reset (so `if(sync)` would be inside try ~ finally).

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
```diff
@@ -321,13 +352,21 @@ private void callDelegateWithParsingError(boolean sync, ApiProxyException expect
     byte[] requestData = new byte[] {0, 1, 2, 3, 4, 5};
     byte[] result = null;
 
+    Logger logger = Logger.getLogger(VmApiProxyDelegate.class.getName());
```
aozarov: Also, did you consider putting it in a private method that would set the Logger and return the previous level?

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
gregw: The logic here is saying that if the log level has not been set to FINE/FINER/FINEST, then raise it to SEVERE, so that WARNINGs are not logged.  This hides the stacks that are generated by the tests, unless somebody sets one of the debug levels indicating they want to see the detail.

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
gregw: OK.  I was making minimal changes, but will add a new try block

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
gregw: OK

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
aozarov: Yes, I understand what it does but I do find it strange. E.g. what if it was INFO, is it OK to just leave it
as INFO at this case?

gregw: If somebody has set logging to INFO, that is kind of the default, so normally they would see INFO, WARNING and SEVERE, which would mean they would see stack traces.   So for this unit test, I'm saying that if the level is INFO (ie it would not log a FINE), then I'm assuming that is default logging and by default I think the stack traces should be suppressed.

If the logging has been set so FINE or above is visible, somebody has asked for details and thus I don't change the logging.

gregw: Note the simplest approach would be to just set the level to SEVERE, but I wanted to allow the option to view the detail logging without changing the code of the unit test.

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
```diff
@@ -194,28 +197,30 @@ private void callDelegateWithOneError(
...
-        result =
-            delegate
-                .makeAsyncCall(
-                    environment, TEST_PACKAGE_NAME, TEST_METHOD_NAME, requestData, apiConfig)
-                .get();
-        fail();
-      } catch (ExecutionException exception) {
-        // ExecutionException is expected, and make sure the cause is expected as well.
-        assertEquals(expectedException.getClass(), exception.getCause().getClass());
+    try (SuppressedLogging l=new SuppressedLogging()) {
```
aozarov: ' = ' the [google-java-format](https://github.com/google/google-java-format) utility didn't fix it for you?

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
```diff
@@ -525,4 +536,20 @@ public void testAPIExceptionWrapping() {
         "The remote RPC to the application server failed for the call barf.d().",
         exception.getMessage());
   }
+  
+  private static class SuppressedLogging implements Closeable {
+    Logger logger = Logger.getLogger(VmApiProxyDelegate.class.getName());
+    Level level = logger.getLevel();
+    
+    SuppressedLogging() {
+      if (!logger.isLoggable(Level.FINE)) {
```
aozarov: I still don't understand this logic of FINE -> SEVERE and rather always set it to SEVERE (per earlier comment),
but if you think that checking for FINE makes sense (e.g. if Level was INFO we would rather keep it as INFO)
then feel free to keep it but a comment explaining this logic would be nice.

aozarov:
Nice! I like this use of Closable. 2 comments and it should be ready for merge.

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
gregw: I'm not using it because it is making a lot of changes to code I have not touched, so it is producing a different result to all the formatting changes you recently committed.    Mostly wrapping stuff, but I've not had time to investigate.

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
gregw: comment added

appengine-managed-runtime/src/test/java/com/google/apphosting/vmruntime/VmApiProxyDelegateTest.java
aozarov: @eamonnmcmanus see comment above. Any reason for the public google-java-format to yield a different result from the internal one?

eamonnmcmanus: It's possible that there are changes to the latest internal version of the formatter that haven't made it to the public version.

PR #261, opened 2016-05-25
Add instructions to deploy with custom built docker image.

yubin154:


yubin154:
I signed it!

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
```diff
@@ -83,6 +83,7 @@
 
   public static final String REQUEST_ENDPOINT = "/rpc_http";
   public static final String REQUEST_STUB_ID = "app-engine-apis";
+  public static final String MEMCACHE_REQUEST_STUB_ID = "app-engine-apis/memcache";
```
aozarov: I think this change should be in a separate PR.

Is that end point already available? Does the local [memcached proxy](https://cloud.google.com/appengine/docs/flexible/java/caching-application-data) already configured to use the same routing?

Also, should that be configurable (so one could revert with a flag if needed)?

yubin154: Second commit reverted. PTAL.

PR #265, opened 2016-05-31
Set all checkstyle checks as error except some Javadoc and few others.

aozarov:
Left as warning few intentional - just discourage but don't force.
Left some of the Javadoc checks such as:
- Missing a Javadoc comment
- Empty line should be followed by <p> tag on the next line.
- First sentence of Javadoc is incomplete (period is missing) or not present.

as there are too many errors with it. We should fix it at some point.


ludoch:
LGTM

ludoch:
Can be tested with this image:
FROM gcr.io/google_appengine/jetty9-compat:06_01_2016_2056
ADD . /app

gregw:
Running google-java-format-1.0-all-deps.jar over the whole code base now produces a few minor line wrapping changes, but it also notices that a lot of our java files are checked in as executables... which is not right.  I'll make another PR for this.

PR #266, opened 2016-05-31
Disable Stackdriver Debugger when running locally

chanseokoh:
Internal bug: 28192592

Do not enable Stackdriver Debugger for local runs.

Please sync this with the master branch too.

### DIFF
--- a/openjdk8/src/main/docker/setup-env.bash
+++ b/openjdk8/src/main/docker/setup-env.bash
@@ -4,13 +4,14 @@ if [[ -n "$ALPN_ENABLE" ]]; then
   ALPN_BOOT="$( /opt/alpn/format-env-appengine-vm.sh )"
 fi
 
-DBG_AGENT="$( RUNTIME_DIR=$JETTY_BASE /opt/cdbg/format-env-appengine-vm.sh )"
-if [[ -n "$DBG_ENABLE" ]]; then
-  if [[ "$GAE_PARTITION" = "dev" ]]; then
+if [[ "$GAE_PARTITION" = "dev" ]]; then
+  if [[ -n "$DBG_ENABLE" ]]; then
     echo "Running locally and DBG_ENABLE is set, enabling standard Java debugger agent"
     DBG_PORT=${DBG_PORT:-5005}
     DBG_AGENT="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=${DBG_PORT}"
   fi
+else
+  DBG_AGENT="$( RUNTIME_DIR=$JETTY_BASE /opt/cdbg/format-env-appengine-vm.sh )"
 fi
 
 PROF_AGENT=


aozarov:
This change looks good to me. @gregw ?

I do wonder about `DBG_ENABLE` as in some cases we document it as is, like [here](https://cloud.google.com/appengine/docs/flexible/java/migrating-an-existing-app#move_your_app_into_the_flexible_environment) but in other places it is mentioned as enabled by default and disabled by [CDBG_DISABLE](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/tree/async-support/jetty9)

gregw:
@aozarov The style that I discovered from cprof when first looking at this (and copied for the alpn initialization) is that there is a `FOO_ENABLE` environment variable in the docker boot script, which if true causes a `FOO` specific `format-env-appengine-vm.sh` script to be called, that then checks for a `FOO_DISABLE` environment variable, which if true makes the call effectively a noop.

So I think this double `ENABLE`/`DISABLE` step is rather strange and must be confusing to document.  Also the cdbg does exactly follow the pattern as the `format-env-appengine-vm.sh` is always called and the  `ENABLE` is only relevant to the dev partition.

I had assumed it was for some backwards compatibility and that a preferred method would eventually be expressed.

Thus if possible, I would rather have just an explicit `ENABLE` or perhaps a `DISABLE` - but whatever is done needs to be the same for alpn, cprof and cdbg.  Failing that  perhaps this change should implement `ENABLE && !DISABLE` logic for both dev and non dev environments?

gregw:
Actually, looking at this again, the logic is even more confusing, because the ENABLED/DISABLED variables are actually only checked for being non null, not a boolean value.   This can be very confusing as CDBG_ENABLE=true and CDGB_ENABLE=false will have the same effect!

At the very least we should make this logic properly check for booleans and be ENABLED && !DISABLED for all three features.    Ideally we could move to just ENABLED logic.

Note that default enable or disabled can easily be supported by the fact that env variables are effectively tri-state: null, true (or 1), false (or 0)

cheers

aozarov:
Yes, that is very confusing. Maybe @ludoch has some more background and also can advise on how we can improve it. I also wondered why we have both `DBG_ENABLE` but then `CDBG_DISABLE` (note the added 'C').

chanseokoh:
Thanks, everyone.

@aozarov Regarding documentation, I am tracking a list of places where it is mentioned that DBG_ENABLE should be set to enable Stackdriver Debugger. I have not fixed these docs yet, since the current live images do need to set DBG_ENABLE. As soon as I confirm that the Debugger is enabled in live images, I will fix all those out-of-sync docs.

For CDBG_DISABLE, I really have no idea about it. What I care about the most is that the Stackdriver Debugger should be enabled by default with no user intervention for all Java runtimes. @gregw You might want to improve the overall ENABLE/DISABLE logic for all the features later. Perhaps later in another PR. Whatever you would do, please make sure the Debugger is enabled by default.

gregw:
I think that we should perhaps accept this PR, as it does a key job and then open another issue to resolve the ENABLE/DISABLE and DBG/CDBG  issues.

PR #269, opened 2016-06-01
Async support google style

gregw:
This PR applies google-java-format-1.0-all-deps.jar to all java files.  Corrects mostly line wraps, but also removes the executable flag on many files.


gregw:
Hmmm this has picked up too many commits.  Closing and will recreate

Issue #267, opened 2016-06-01
Fix the env settings for enable/disable debugger


aozarov:
Currently they are set as `DBG_ENABLED` and `CDBG_DISABLE`. We should make their names consistent. 

Also, flag are set to be on or off based on being an empty string or not.
We should consider applying a boolean value check.

For more info see #266

gregw:
To plan this change, we need to work out external dependencies:

- CDBG_DISABLE is used within `/opt/cdbg/format-env-appengine-vm.sh` which is imported from [cdbg_java_agent.tar.gz](https://storage.googleapis.com/cloud-debugger/appengine-java/current/cdbg_java_agent.tar.gz)
- DBG_ENABLE is used only with scripts from this project, but is listed in [documentation](https://cloud.google.com/appengine/docs/flexible/java/migrating-an-existing-app#move_your_app_into_the_flexible_environment), which can be changed in the core doco, but not in [many links](https://www.google.com.au/webhp?q=DBG_ENABLE+appengine).
- the current `setup-env.bash` assumes that a `/opt/cprof/format-env-appengine-vm.sh` file exists that is called if CPROF_ENABLE is set, but that script is no longer in the import of [cloud_profiler_java_agent.tar.gz](https://storage.googleapis.com/cloud-profiler/appengine-java/current/cloud_profiler_java_agent.tar.gz). So it appears that this dependency has changed and we've not noticed.

chanseokoh:
About CDBG_DISABLE:

Stackdriver Debugger is enabled by default on all runtimes, the only exception being the Java runtimes. This exception in Java seems to be due to some unfortunate mishap, and we fixed it a while ago after noticing it (#219). Note that CDBG_DISABLE is inherent to the Stackdriver Debugger design. Because the debugger is enabled by default, people needed a way to disable it for various reasons, so it was decided to use this variable. The name of the variable cannot be changed as this variable is being used across several internal systems.

aozarov:
@chanseokoh in that case shouldn't we change the check in this [line](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/async-support/openjdk8/src/main/docker/setup-env.bash#L8) to -z "${CDBG_DISABLE}" ?

If so, we should also update the doc which I think you suggested in #226 that you will do?

chanseokoh:
[...]

PR #276, opened 2016-06-04
DBG_ENABLE #267

gregw:
DBG_ENABLE cleanup for #267 

This uses case insensitive booleans for both ALPN_ENABLE and DBG_ENABLE

If the booleans are not set, then ALPN_ENABLE defaults to false and DBG_ENABLE defaults to true, unless CDBG_DISABLE is set to non null.

I'm not sure we want to defer to the CDBG_DISABLE in that way?


openjdk8/src/main/docker/setup-env.bash
```diff
@@ -1,18 +1,27 @@
...
 ALPN_BOOT=
-if [[ -n "$ALPN_ENABLE" ]]; then
+if isTrue "${ALPN_ENABLE:=false}" ; then
   ALPN_BOOT="$( /opt/alpn/format-env-appengine-vm.sh )"
 fi
 
 DBG_AGENT=
-if [[ "$GAE_PARTITION" = "dev" ]]; then
-  if [[ -n "$DBG_ENABLE" ]]; then
+if isTrue "${DBG_ENABLE:=$( if [[ -z ${CDBG_DISABLE} ]] ; then echo true; else echo false ; fi )}" ; then
```
aozarov: Fine as is but do you really think we should also consider CDBG_DISABLE as a fallback to DBG_ENABLE?
From the user perspective I thought we said that we only document and consider  DBG_ENABLE (and then
manipulate the other "internal" variables" based on it, as you already do).

gregw: I'm 50:50 on this.  I can see that it is a good backwards compatibility fallback... but then I can't find any documentation about it either... other than the code in the cdbg script.   Although I can imagine somebody looking at that script and then being driven crazy setting CDBG_DISABLE and wondering why it is not disabled.

Note that even if we don't respect it, we still have to unset it in our script, so we can't totally remove handling for CDBG_DISABLE.  If we have to code for it, then why not respect it?    Hmmm think I'm talking myself to 60:40 keeping it.

But its OK either way.  your call!

aozarov: Ok, lets leave it but not document it.

aozarov:
Do you mind to include the doc change for this script change in this PR?

gregw:
@aozarov I've updated the `README.md` of openjdk8 and jetty9, but the `README.md` for jetty-9-compat is woefully inadequate and also on the base rather than the image.  I'll open another issue for that.

PR #278, opened 2016-06-04
Update README.md

aozarov:


README.md
```diff
@@ -12,7 +12,7 @@ the image, you need to build it with either a local docker installation or confi
 This will create the following docker images:
  * openjdk8:8-jre
  * jetty9:9.3.x
- * jetty9-compatible:1.9.x
+ * jetty9-compat
```
gregw: should this include a label like the other two?   or perhaps remove the other labels? or perhaps make them variables like `${jetty.version}` and `${appengine.api.version}`?

aozarov: If I understand correctly the context of this doc, it seems to me related to a local build (after `mvn clean install` and before using the result in a local Dockerfile for the app, which we suggest to use the `latest` label). If that is the case, then I suggest to remove the label, as the label is tied to more parts (e.g. **async**-1.9.38-**SNAPSHOT**).

gregw: OK by me.

PR #286, opened 2016-06-09
Fix #243 Configure Heap and GC options

### LINKED ISSUE #243: Review memory allocation and GC configuration
The jetty9-compat [start script](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/async-support/jetty9-compat/src/main/docker/jetty_run.sh) contains a calculation of the allowed HEAP_SIZE which is used to start java with `-Xms${HEAP_SIZE} -Xmx${HEAP_SIZE}`

We need to review if this calculation is a good one.  Also we should consider what other GC options/tuning should be done.

gregw:
Fix #243 Configure Heap and GC options

This is a bit bigger refactor than was discussed.  I've moved the Java HEAP, GC, GC log, etc all into the setup-env.bash script so that it is in a common point for all images.   I've JAVA_OPTS from their built from many parts, each that can be overridden or defaulted, so the user can change parts without taking over all of JAVA_OPTS.

Looks like I need to do a merge to get the other updated PRs before this one is ready to merge... stand by...


jetty9/README.md
```diff
@@ -88,23 +88,35 @@ The entry point for the image is [docker-entrypoint.bash](https://github.com/Goo
...
 
-| Feature              | directory    | Enable/Disable  | Default | JVM args      |
-|----------------------|--------------|-----------------|---------|---------------|
-| Stackdriver Debugger | /opt/cdbg/   | $DBG_ENABLE     | true    | $DBG_AGENT    |
-| Temporary file       |              | $TMPDIR         |         | $SET_TMP      |
-| Java options         |              | $JAVA_OPTS      |         | $JAVA_OPTS    |
+| Feature              | directory  | Enable/Disable  | Default            | JVM args           |
+|----------------------|------------|-----------------|--------------------|--------------------|
+| Stackdriver Debugger | /opt/cdbg/ | $DBG_ENABLE     | true               | $DBG_AGENT         |
+| Temporary file       |            | $TMPDIR         |                    | $JAVA_TMP_OPTS     |
```
aozarov: s/file/directory/ ?

jetty9/README.md
```diff
@@ -88,23 +88,35 @@ The entry point for the image is [docker-entrypoint.bash](https://github.com/Goo
...
+| Temporary file       |            | $HEAD_SIZE      | from /proc/meminfo | $JAVA_HEAP_OPTS    |
```
aozarov: Jave Heap?

openjdk8/README.md
```diff
@@ -38,15 +38,30 @@ The entry point for the openjdk8 image is [docker-entrypoint.bash](https://githu
...
 
-| Feature              | directory    | Enable/Disable  | Default | JVM args      |
-|----------------------|--------------|-----------------|---------|---------------|
-| Stackdriver Debugger | /opt/cdbg/   | $DBG_ENABLE     | true    | $DBG_AGENT    |
-| Temporary file       |              | $TMPDIR         |         | $SET_TMP      |
-| Java options         |              | $JAVA_OPTS      |         | $JAVA_OPTS    |
+| Feature              | directory  | Enable/Disable  | Default            | JVM args           |
+|----------------------|------------|-----------------|--------------------|--------------------|
+| Stackdriver Debugger | /opt/cdbg/ | $DBG_ENABLE     | true               | $DBG_AGENT         |
+| Temporary file       |            | $TMPDIR         |                    | $JAVA_TMP_OPTS     |
```
aozarov: see comments above

openjdk8/README.md
```diff
@@ -38,15 +38,30 @@ The entry point for the openjdk8 image is [docker-entrypoint.bash](https://githu
...
+| Temporary file       |            | $HEAD_SIZE      | from /proc/meminfo | $JAVA_HEAP_OPTS    |
```
aozarov: HEAP_SIZE ?

openjdk8/README.md
```diff
@@ -38,15 +38,30 @@ The entry point for the openjdk8 image is [docker-entrypoint.bash](https://githu
 
 If the default command (java) is used, then the entry point sources the [setup-env.bash](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/master/openjdk8/src/main/docker/setup-env.bash), which looks for supported features to be enabled and/or configured.  The following table indicates the environment variables that may be used to enable/disable features, any default values if they are not set and the environment variable that holds the resulting JVM arguments that actually enable the feature: 
 
-| Feature              | directory    | Enable/Disable  | Default | JVM args      |
-|----------------------|--------------|-----------------|---------|---------------|
-| Stackdriver Debugger | /opt/cdbg/   | $DBG_ENABLE     | true    | $DBG_AGENT    |
-| Temporary file       |              | $TMPDIR         |         | $SET_TMP      |
-| Java options         |              | $JAVA_OPTS      |         | $JAVA_OPTS    |
+| Feature              | directory  | Enable/Disable  | Default            | JVM args           |
```
aozarov: enable/disable is confusing now when we have TMP_DIR and the other options.

Did you consider instead to have 2 tables. first table as before to enable/disable features and the second table to specify the options that could be configured by users (and what their default value is)?

openjdk8/src/main/docker/setup-env.bash
```diff
@@ -19,8 +19,13 @@ if isTrue "${DBG_ENABLE:=$( if [[ -z ${CDBG_DISABLE} ]] ; then echo true; else e
   fi
 fi
 
-SET_TMP=
-if [[ -n "${TMPDIR}" ]]; then
-  SET_TMP="-Djava.io.tmpdir=$TMPDIR"
-fi
+# Setup default Java Options
+: ${JAVA_TMP_OPTS:=$( if [[ -z ${TMPDIR} ]]; then echo ""; else echo "-Djava.io.tmpdir=$TMPDIR"; fi)}
```
aozarov: This may need to change if you accept my comment above.

jetty9/README.md
gregw: Yes the directory column doesn't really mean much any more.  I think it should just be deleted.

openjdk8/README.md
gregw: It is actually a graph of env variable dependencies and users can set any of them.
I try an alternate format that shows that better.
stand by

openjdk8/src/main/docker/setup-env.bash
gregw: @aozarov How does the table look now?

jetty9/README.md
```diff
@@ -86,25 +86,40 @@ Once you have this configuration, you can use the Google Cloud SDK to deploy thi
...
-|----------------------|--------------|-----------------|---------|---------------|
-| Stackdriver Debugger | /opt/cdbg/   | $DBG_ENABLE     | true    | $DBG_AGENT    |
-| Temporary file       |              | $TMPDIR         |         | $SET_TMP      |
-| Java options         |              | $JAVA_OPTS      |         | $JAVA_OPTS    |
+If the default command (java) is used, then the entry point sources the [setup-env.bash](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/master/openjdk8/src/main/docker/setup-env.bash), which looks for supported features to be enabled and/or configured.  The following table indicates the environment variables that may be used to enable/disable/configure features, any default values if they are not set: 
+
+|Env Var           | Description         | Type     | Default                               |
+|------------------|---------------------|----------|---------------------------------------|
+|`DBG_ENABLE`      | Stackdriver Debugger| Boolean  | `true`                                |
+|`TMPDIR`          | Temporary Directory | dirname  |                                       |
```
aozarov: Nit, should we use Capitalize all types (Boolean Size, Dirname, Filename)? or use lowercase for all?

aozarov:
Nice change! feel free to merge with or without my nit.

Issue #245, opened 2016-05-20
Origin port not set for jetty-compat-base running locally


janbartel:
The origin port for requests is not set when running the jetty9-compat-base locally (ie outside a google firewall).

This means that response.sendRedirect(location) produces an encoded location missing the port. Eg if response.sendRedirect("/foo") is called, the location header produced is "http://localhost/foo" instead of "http://localhost:8080/foo".

ludoch:
I guess we can close this one?

PR #287, opened 2016-06-09
Async support 245

janbartel:


gregw:
This PR removes the duplication of the request customization code.  It was done in a request customizer for Jetty-9, but as part of the VmRuntimeWebAppContext in jetty9-compat.     Jan's PR makes both images use the request customizer , but in jetty9-compat the configuration in XML includes the env variable or system property that were previously checked in code.

It fixes issue #245 because the request customizer version of the code is tri-state: on/off/unset with unset being for local invocation and the request authority is not changed.

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/JettyRunner.java
```diff
@@ -166,6 +166,10 @@ public void doStart() throws Exception {
       httpConfig.setSendServerVersion(true);
       httpConfig.setSendDateHeader(false);
       httpConfig.setDelayDispatchUntilContent(false);
+      GoogleRequestCustomizer requestCustomizer = new com.google.apphosting.jetty9.GoogleRequestCustomizer();
```
aozarov: Any reason for not importing it?

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/JettyRunner.java
```diff
@@ -166,6 +166,10 @@ public void doStart() throws Exception {
...
+      requestCustomizer.setHttpPort(port);
+      requestCustomizer.setHttpsPort(Integer.parseInt("443"));
```
aozarov: Any reason to create the integer from a string?

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/JettyRunner.java
```diff
@@ -166,9 +167,8 @@ public void doStart() throws Exception {
       httpConfig.setSendServerVersion(true);
       httpConfig.setSendDateHeader(false);
       httpConfig.setDelayDispatchUntilContent(false);
-      GoogleRequestCustomizer requestCustomizer = new com.google.apphosting.jetty9.GoogleRequestCustomizer();
-      requestCustomizer.setHttpPort(port);
-      requestCustomizer.setHttpsPort(Integer.parseInt("443"));
+      GoogleRequestCustomizer requestCustomizer =
+          new GoogleRequestCustomizer(port, Integer.parseInt("443"));
```
aozarov: I still don't understand why you need the `Integer.parseInt("443")` conversion rather than just passing
the literal `443`...

janbartel: Sorry didn't see your previous comment before my commit. Fixed now.

PR #288, opened 2016-06-09
Fixes #281 - Enforcer Rules update

### LINKED ISSUE #281: Ensure the maven enforcer rule checks for Maven 3.3+
As a result of Issue #262 it was identified that the Maven version should be a bit more strict in the enforcer rule.

The maven enforcer needs to make sure that the maven version tests minor version as well as the major version.

joakime:
+ maven-enforcer-plugin now runs on `validate phase
+ maven version enforcement is 3.3+ now

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,29 +107,6 @@
   <build>
     <pluginManagement>
       <plugins>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-enforcer-plugin</artifactId>
-          <version>1.4.1</version>
-          <executions>
-            <execution>
-              <id>enforce-maven</id>
-              <goals>
-                <goal>enforce</goal>
-              </goals>
-              <configuration>
-                <rules>
-                  <requireMavenVersion>
-                    <version>[3.0,)</version>
-                  </requireMavenVersion>
-                  <requireJavaVersion>
-                    <version>[1.8,)</version>
-                  </requireJavaVersion>
-                </rules>
-              </configuration>
-            </execution>
-          </executions>
-        </plugin>
 
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


ludoch:
Does not know how to tell circleci to use a newer Maven version... :-(

joakime:
This PR makes the enforcer rules run everywhere now (btw).
Prior to this PR the enforcer rules didn't run.

joakime:
I can't see a way to tell travis to use maven 3.3.x either.

joakime:
According to CircleCI, the only way to upgrade the maven version is to upgrade the build-image.

Unfortunately, they only have Ubuntu 14.04, which is stuck at maven 3.2.5.
Wish they would offer Fedora, those images are at least up to date with their build tools.

When we have to start working with JDK9 (which the Jetty project is already doing in Jetty 9.4.x), then the ability to manage the JDK and Maven versions will be critical for any CI choice.

Still haven't found a TravisCI solution for maven.

aozarov:
I don't think Travis support it as a [configuration](https://github.com/travis-ci/enterprise-installation/issues/14) however, I think that could be done by installing it via [apt-get](https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon) or getting it in the [before-install](https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle) stage.

However, because that is not trivial in both cases do we really need it?

joakime:
Just curious, can a paid circleci instance use custom images?

Might want to start evaluating better CI solutions that we can control better.
The travis and circleci efforts will be short lived in any case for this project.

joakime:
rebuild pls?

meltsufin:
I started the rebuild, but it failed again. Also, notice that there are merge conflicts.

gregw:
no longer required

Issue #259, opened 2016-05-25
 Remove from the jetty image (or hide from the application) the AE API jar


aozarov:
Currently we *remove* the AE API jar provided by the application and use the same one that is bundled with Jetty (needed for session management). 

I think ideally we should hide (different class-loader) the AE API jar that is bundled with Jetty and keep AE API jar that is bundled with the application. 

I also think that it would be nice if we, at least, warn during deployment time if that jar is missing from compat mode.

Also, it would be nice if dev appsesrver would be consistent with production and not expose the AE API jar if not provided by the user.

gregw:
This can be achieved with the jetty server class mechanism.      We can manually update the list of hidden packages/classes, but a better approach would be to use maven to automatically generate the list.

aozarov:
I think this issue is also related to #253

gregw:
@aozarov Just to be clear,  the "remove the AE API jar provided by the application" is done by the SDK?   In which case if we hide the AE API jar from the webapp we will break old versions of the SDK.

Instead, we can be dynamic and examine the webapp for a jar providing com/google/appengine/api to trigger specific behaviour, for which I see two options:

1. We always hide com.google.* packages (and their dependencies) from the webapp.  If the webapp does not have it's own com.google.appengine.api then we add the api jar (with all its bundled dependencies and repackaging) to the webapp classloader.
2. We only hide com.google.* from the webapp if it has it's own com.google.appengine.api, otherwise we allow it to see the classes from the server classloader

Option 1 is simplest and more robust for changes to the contents of the jar, but will result in more memory usage as the api jar is loaded twice.  

Option 2 saves memory (but only while the SDK is removing the API jar) but is more complex choosing exactly what packages to expose is not obvious and may be subject to change.

aozarov:
> @aozarov Just to be clear, the "remove the AE API jar provided by the application" is done by the SDK? In which case if we hide the AE API jar from the webapp we will break old versions of the SDK.

Yes, this is done (according to @ludoch) by the SDK (I think in the staging phase).
Indeed that means it is a breaking change but I think we should do that. @ludoch may disagree.

If possible to restrict it only for flex deployments and therefore minimize the backward compatibility
issues I would take that. One option is to change the staging to stop removing the api jar only for flex
and making this change to the runtime only for flex (by either saying async-support branch is only
for flex or changing the class-loading behavour based on flex or not). @ludoch any thoughts about it?

[...]

PR #290, opened 2016-06-15
Issue #259 Hide GAE API from webapp classpath

gregw:
Handles #259 bu creating AppengineApiConfiguration to manage the classpath that is exposed and/or protected from the webapplication.    If the GAE API is not available in the webapp, then a warning is logged and the API is added to the classpath.

Currently javax.activation and javax.mail are treated as shared classes and the containers version is made available. If these are not listed, then even if they are in the added API jar, errors occur, which needs to be investigated.

Currently a specific list of org.apache packages are hidden.   A better approach may be to hide all org.apache classes and selectively reveals those needed for JSP, JSTL and EL support?

### DIFF
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
@@ -24,6 +24,9 @@ import java.util.logging.Level;
 
 public class AppengineApiConfiguration extends AbstractConfiguration {
 
+  // A class to check if the GAE API is available.
+  public final static String GAE_CHECK_CLASS = "com.google.appengine.api.ThreadManager";
+  
   // Hide the all container classes from the webapplication
   // TODO update to use '.' when supported by Jetty
   private static final String[] SERVER_CLASSES = {
@@ -34,6 +37,7 @@ public class AppengineApiConfiguration extends AbstractConfiguration {
   };
 
   private static final String[] SHARED_CLASSES = {
+    // Expose the GAE API classes
     "com.google.appengine.api.LifecycleManager",
     "com.google.apphosting.api.ApiProxy",
     "com.google.apphosting.api.ApiStats",
@@ -64,9 +68,9 @@ public class AppengineApiConfiguration extends AbstractConfiguration {
       context.addServerClass(systemClass);
     }
     for (String gaeClass : SHARED_CLASSES) {
-      // Don't hide GAE v1 shared classes
+      // Don't hide shared classes
       context.prependServerClass("-" + gaeClass);
-      // Don't GAE v1 shared classes to be replaced by webapp
+      // Don't allow shared classes to be replaced by webapp
       context.addSystemClass(gaeClass);
     }
   }
@@ -75,11 +79,11 @@ public class AppengineApiConfiguration extends AbstractConfiguration {
     ClassLoader loader = context.getClassLoader();
     try {
       // Test if the appengine api is available
-      loader.loadClass("com.google.appengine.api.ThreadManager");
+      loader.loadClass(GAE_CHECK_CLASS);
     } catch (Exception ex) {
       VmRuntimeWebAppContext.logger.log(Level.FINE, "No appengined API", ex);
       VmRuntimeWebAppContext.logger.log(
-          Level.WARNING, "No appengined API including in WEB-INF/lib! Please update your SDK!");
+          Level.WARNING, "No appengined API jar including in WEB-INF/lib! Please update your SDK!");
 
       // The appengine API is not available so we will add it and it's dependencies
       Resource providedApi =
@@ -98,7 +102,7 @@ public class AppengineApiConfiguration extends AbstractConfiguration {
       }
 
       // Ensure the API can now be loaded
-      loader.loadClass("com.google.appengine.api.ThreadManager");
+      loader.loadClass(GAE_CHECK_CLASS);
     }
   }
 }
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.apphosting.vmruntime.jetty9;
+
+import org.eclipse.jetty.util.URIUtil;
+import org.eclipse.jetty.util.resource.Resource;
+import org.eclipse.jetty.webapp.AbstractConfiguration;
+import org.eclipse.jetty.webapp.WebAppClassLoader;
+import org.eclipse.jetty.webapp.WebAppContext;
+
+import java.util.Arrays;
+import java.util.logging.Level;
+
+public class AppengineApiConfiguration extends AbstractConfiguration {
+
+  // Hide the server implementation classes from the webapplication
+  // TODO Regularly review this list; or automate generation; or change to whitelist
+  private static final String[] SERVER_CLASSES = {
+    "com.google.",
+    "javax.cache.",
+    "org.apache.commons.",
+    "org.apache.geronimo.",
+    "org.apache.http.",
+    "mozilla."
+  };
+
+  private static final String[] GAE_STANDARD_V1_SHAREED_CLASSES = {
+    "com.google.appengine.api.LifecycleManager",
+    "com.google.apphosting.api.ApiProxy",
+    "com.google.apphosting.api.ApiStats",
+    "com.google.apphosting.api.CloudTrace",
+    "com.google.apphosting.api.CloudTraceContext",
+    "com.google.apphosting.api.DeadlineExceededException",
+    "com.google.apphosting.runtime.SessionData",
+    "com.google.apphosting.runtime.UncatchableError",
+    
+    // TODO review if these should be shared or provided?
+    "javax.activation.",
+    "javax.mail.",
+  };
+
+  @Override
+  public void preConfigure(WebAppContext context) throws Exception {
+    for (String systemClass : SERVER_CLASSES) {
+      context.addServerClass(systemClass);
+    }
+    for (String gaeClass : GAE_STANDARD_V1_SHAREED_CLASSES) {
+      // Don't hide GAE v1 shared classes
+      context.prependServerClass("-" + gaeClass);
+      // Don't GAE v1 shared classes to be replaced by webapp
+      context.addSystemClass(gaeClass);
+    }
+  }
+
+  public void configure(WebAppContext context) throws Exception {
+    ClassLoader loader = context.getClassLoader();
+    try {
+      // Test if the appengine api is available
+      loader.loadClass("com.google.appengine.api.ThreadManager");
+    } catch (Exception ex) {
+      VmRuntimeWebAppContext.logger.log(Level.FINE, "No appengined API", ex);
+      VmRuntimeWebAppContext.logger.log(
+          Level.WARNING, "No appengined API including in WEB-INF/lib! Please update your SDK!");
+
+      // The appengine API is not available so we will add it and it's dependencies
+      Resource providedApi =
+          Resource.newResource(
+              URIUtil.addPaths(System.getProperty("jetty.base"), "/lib/gae/provided-api/"));
+
+      if (providedApi != null) {
+        String[] list = providedApi.list();
+        if (list != null) {
+          WebAppClassLoader wloader = (WebAppClassLoader) loader;
+          for (String jar : list) {
+            wloader.addClassPath(providedApi.addPath(jar));
+            VmRuntimeWebAppContext.logger.log(Level.INFO, "Added " + jar + " to webapp classpath");
+          }
+        }
+      }
+
+      // Ensure the API can now be loaded
+      loader.loadClass("com.google.appengine.api.ThreadManager");
+    }
+  }
+}
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
@@ -38,6 +38,10 @@ public class AppengineApiConfiguration extends AbstractConfiguration {
     "mozilla."
   };
 
+  // Classes & Packages to be shared from the containers classloader
+  // to the webapp.  Classes included here will be marked both as 
+  // exclusions from serverclasses (not hidden from webapp) and inclusion
+  // to the systemclasses (cannot be overriden by the webapp)
   private static final String[] SHARED_CLASSES = {
     // Expose the GAE API classes
     "com.google.appengine.api.LifecycleManager",
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
@@ -25,14 +25,14 @@ import java.util.logging.Level;
 public class AppengineApiConfiguration extends AbstractConfiguration {
 
   // A class to check if the GAE API is available.
-  public final static String GAE_CHECK_CLASS = "com.google.appengine.api.ThreadManager";
-  
+  public static final String GAE_CHECK_CLASS = "com.google.appengine.api.ThreadManager";
+
   // Hide the all container classes from the webapplication
   // TODO update to use '.' when supported by Jetty
   private static final String[] SERVER_CLASSES = {
-    "com.",
-    "javax.",
-    "org.",
+    "com.", 
+    "javax.", 
+    "org.", 
     "mozilla."
   };
 
@@ -46,20 +46,20 @@ public class AppengineApiConfiguration extends AbstractConfiguration {
     "com.google.apphosting.api.DeadlineExceededException",
     "com.google.apphosting.runtime.SessionData",
     "com.google.apphosting.runtime.UncatchableError",
-    
+
     // Expose the standard APIs that are provided
     "javax.servlet.",
     "javax.el.",
     "javax.annotation.",
     "javax.activation.", // TODO Review
     "javax.mail.", // TODO Review
-    
+
     // Expose classes needed for JSP and JSTL
     "org.apache.jasper.runtime.",
     "org.apache.jasper.JasperException",
     "org.apache.el.ExpressionFactoryImpl",
     "org.apache.tomcat.InstanceManager",
-    "org.apache.taglibs.",    
+    "org.apache.taglibs.",
   };
 
   @Override
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
@@ -1,15 +1,17 @@
 /*
  * Copyright 2016 Google Inc. All Rights Reserved.
  *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
- * in compliance with the License. You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *       http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software distributed under the License
- * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied. See the License for the specific language governing permissions and limitations under
- * the License.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 
 package com.google.apphosting.vmruntime.jetty9;
@@ -112,8 +114,13 @@ public class AppengineApiConfiguration extends AbstractConfiguration {
       loader.loadClass(GAE_CHECK_CLASS);
     } catch (Exception ex) {
       VmRuntimeWebAppContext.logger.log(Level.FINE, "No appengined API", ex);
-      VmRuntimeWebAppContext.logger.log(
-          Level.WARNING, "No appengined API jar including in WEB-INF/lib! Please update your SDK!");
+      if (VmRuntimeWebAppContext.logger.isLoggable(Level.FINE)) {
+        VmRuntimeWebAppContext.logger.log(Level.WARNING,
+            "No appengine API jar included in WEB-INF/lib! Please update your SDK!", ex);
+      } else {
+        VmRuntimeWebAppContext.logger.log(Level.WARNING,
+            "No appengine API jar included in WEB-INF/lib! Please update your SDK!");
+      }
 
       // The appengine API is not available so we will add it and it's dependencies
       Resource providedApi =
--- a/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
+++ b/jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
@@ -47,13 +47,42 @@ public class AppengineApiConfiguration extends AbstractConfiguration {
     "com.google.apphosting.runtime.SessionData",
     "com.google.apphosting.runtime.UncatchableError",
 
-    // Expose the standard APIs that are provided
+    // Expose the standard APIs that are provided by the container
     "javax.servlet.",
     "javax.el.",
-    "javax.annotation.",
-    "javax.activation.", // TODO Review
     "javax.mail.", // TODO Review
-    "javax.imageio.", // TODO Review: Required by guestbook example
+    
+    // Expose the standard APIs that are provided by the JVM
+    "com.oracle.",
+    "com.sun.",
+    "javax.accessibility.",
+    "javax.activation.",
+    "javax.activity.",
+    "javax.annotation.",
+    "javax.imageio.",
+    "javax.jws.",
+    "javax.lang.model.",
+    "javax.management.",
+    "javax.naming.",
+    "javax.net.",
+    "javax.print.",
+    "javax.rmi.",
+    "javax.script.",
+    "javax.security.",
+    "javax.smartcardio.",
+    "javax.sound.",
+    "javax.sql.",
+    "javax.swing.",
+    "javax.tools.",
+    "javax.transaction.",
+    "javax.xml.",
+    "jdk.",
+    "org.ietf.jgss.",
+    "org.jcp.xml.dsig.internal.",
+    "org.omg.",
+    "org.w3c.dom.",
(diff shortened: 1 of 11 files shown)


gregw:
This passes the integration tests and has been tested against the guestbook stage3.  However I think it probably needs wider testing before merging.    An image is available for testing against at `gcr.io/jetty9-work/jetty9-compat:gregw-259`

gregw:
@aozarov I understand you are very busy this week and may not be able to review this PR in detail.   Currently I don't think it is really ready to merge anyway, but I just created it to get any feedback that you can give on the approach while I continue to improve.

The need to expose `javax.activation` and `javax.mail` is strange as they are provided in the API jar that is added to the context classpath.  So I need to investigate more why the context version is not acceptable.

Also need more applications to test again this.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+    "com.google.appengine.api.LifecycleManager",
+    "com.google.apphosting.api.ApiProxy",
+    "com.google.apphosting.api.ApiStats",
+    "com.google.apphosting.api.CloudTrace",
+    "com.google.apphosting.api.CloudTraceContext",
+    "com.google.apphosting.api.DeadlineExceededException",
+    "com.google.apphosting.runtime.SessionData",
+    "com.google.apphosting.runtime.UncatchableError",
+    
+    // TODO review if these should be shared or provided?
```
aozarov: lets deal with these 2 in a separate issue. Ideally should not be included in the SDK and provided by the user based on need.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+    "javax.cache.",
+    "org.apache.commons.",
+    "org.apache.geronimo.",
+    "org.apache.http.",
+    "mozilla."
+  };
+
+  private static final String[] GAE_STANDARD_V1_SHAREED_CLASSES = {
+    "com.google.appengine.api.LifecycleManager",
+    "com.google.apphosting.api.ApiProxy",
```
aozarov: Is this a prefix and will also work for the [inner classes](https://cloud.google.com/appengine/docs/java/javadoc/com/google/apphosting/api/package-summary)?

gregw: yes, matching a class FooBar also matches any innerclasses like FooBar$InnerMe

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+import org.eclipse.jetty.webapp.WebAppClassLoader;
+import org.eclipse.jetty.webapp.WebAppContext;
+
+import java.util.Arrays;
+import java.util.logging.Level;
+
+public class AppengineApiConfiguration extends AbstractConfiguration {
+
+  // Hide the server implementation classes from the webapplication
+  // TODO Regularly review this list; or automate generation; or change to whitelist
```
aozarov: Shouldn't it be all classes that are not part of the application's WAR (lib/classes) except of the shared ones?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
gregw: Ok.  As is they are shared and that works for the very limited tests that I have.   I'd still like to work out why providing them didn't actually work, but as I'm struggling for time to look into that, best if it does not hold up this PR.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
gregw: Yes.  

But because the the history where all jars placed on the server classpath are exposed unless explicitly hidden, we don't have a mechanism to hide all and then reveal specific.

I'll raise a jetty issue to make sure we have a matchall pattern in the next release.  Potentially a pattern of "." might match everything now, but it is a poor match-all pattern.  Stand by I will test that...

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+    "com.google.apphosting.runtime.UncatchableError",
+    
+    // TODO review if these should be shared or provided?
+    "javax.activation.",
+    "javax.mail.",
+  };
+
+  @Override
+  public void preConfigure(WebAppContext context) throws Exception {
+    for (String systemClass : SERVER_CLASSES) {
```
aozarov: does not look like anything in this method will actually throw an exception.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+
+      // The appengine API is not available so we will add it and it's dependencies
+      Resource providedApi =
+          Resource.newResource(
+              URIUtil.addPaths(System.getProperty("jetty.base"), "/lib/gae/provided-api/"));
+
+      if (providedApi != null) {
+        String[] list = providedApi.list();
+        if (list != null) {
+          WebAppClassLoader wloader = (WebAppClassLoader) loader;
```
aozarov: Any reason for not using a covariant return type in WebAppContext#getClassLoader to return WebAppClassLoader?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
gregw: No a pattern of "." does not match all packages and classes.    
While I don't like it as a match all pattern, I think it is the most consistent with the style of classpath pattern matching we have.  So I've opened a [jetty issue](https://github.com/eclipse/jetty.project/issues/673) and will fix that in the next release.

Shall I change this PR to blocking "com.","org." for now?

Note that we would then need to explicitly reveal the org.apache classes needed for JSP and EL.

aozarov: That would be good if you can make it work.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -271,6 +282,12 @@ public VmRuntimeWebAppContext() {
     ApiProxy.setDelegate(new VmApiProxyDelegate());
   }
 
+  @Override
+  public void configure() throws Exception {
+    // TODO Auto-generated method stub
```
aozarov: remove comment, but also why override?

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/JettyRunner.java
```diff
@@ -192,7 +186,12 @@ public void doStart() throws Exception {
       // configuration from root.xml
       final VmRuntimeWebAppContext context = new VmRuntimeWebAppContext();
       context.setContextPath("/");
-      context.setConfigurationClasses(preconfigurationClasses);
+
+      // remove annotations
+      context.setConfigurationClasses(
+          Arrays.stream(context.getConfigurationClasses())
+              .filter(n -> !n.contains("AnnotationConfiguration"))
```
aozarov: Maybe leave the comment "// next one is way too slow for unit testing:" here?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -185,16 +181,38 @@ protected void doStart() throws Exception {
...
-    datastoreService = getDatastoreService();
-
     if (quickstartWebXml == null) {
       addEventListener(new ContextListener());
     } else {
       getMetaData()
           .addDescriptorProcessor(preconfigProcessor = new PreconfigureDescriptorProcessor());
     }
 
+    // DatastoreServiceFactory.getDatastoreService();
```
aozarov: You can probably remove the related import, however can you explain why is that necessary?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
gregw: WebAppContext does not require that a setClassLoader uses a WebAppClassLoader and any calls that attempt to use the specific API are protected by instanceof calls internally.        Now this is a very rarely used capability, but it has been done enough that we can't use a covariant  return type.

In this case, we are assuming that the classloader is a WebAppClassLoader, which is safe because we control the entire environment.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
gregw: The previous code was fetching the datastoreService to "facilitate testing"? but then it is not actually used.
So I think all the code and comments and imports can be removed.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
gregw: Indeed - will delete method

gregw:
@aozarov I'll be pushing a few changes to this branch, but hold off reviewing/testing until I call you out again.

gregw:
@aozarov This PR now runs the IT tests and I can deploy stage3 of the guestbook on it.    The guestbook required that I also expose `javax.imageio`, so my expectation is that there may be other javax packages that should be exposed.

I can see about 25 javax packages that are in the JVMs rt.jar, 1 in jce.jar.   Should I white list these or should we revert to a black list for javax?

Also, we need to decide if we wish to hide other JVM packages like: `com.sun.`, `sun.`, `javafx`, `com.oracle.`   ?

aozarov:
@gregw I don't think we should hide JVM packages that are included with the Java runtime.

gregw:
@aozarov agreed.  I've white listed all the packages provided by the java runtime.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
aozarov: Nit, I think this license is not aligned correctly.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+  private static final String[] SERVER_CLASSES = {
+    "com.", 
+    "javax.", 
+    "org.", 
+    "mozilla."
+  };
+
+  private static final String[] SHARED_CLASSES = {
+    // Expose the GAE API classes
+    "com.google.appengine.api.LifecycleManager",
```
aozarov: Will these classes always be picked from the server classes when also bundled with the application?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+    "com.google.apphosting.api.ApiProxy",
+    "com.google.apphosting.api.ApiStats",
+    "com.google.apphosting.api.CloudTrace",
+    "com.google.apphosting.api.CloudTraceContext",
+    "com.google.apphosting.api.DeadlineExceededException",
+    "com.google.apphosting.runtime.SessionData",
+    "com.google.apphosting.runtime.UncatchableError",
+
+    // Expose the standard APIs that are provided by the container
+    "javax.servlet.",
```
aozarov: are there any javax.* packages that are not part of the JDK (such as javax.servlet) that
we bundle with jetty and don't want to expose? Oh, one is probably javax.cache.

So, I wonder if it would not be simpler just to mention the specific javax packages in the SERVER_CLASSES and then I assume you don't need to specify them individually in SHARED_CLASSES?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+  @Override
+  public void preConfigure(WebAppContext context) {
+    for (String systemClass : SERVER_CLASSES) {
+      context.addServerClass(systemClass);
+    }
+    for (String gaeClass : SHARED_CLASSES) {
+      // Don't hide shared classes
+      context.prependServerClass("-" + gaeClass);
+      // Don't allow shared classes to be replaced by webapp
+      context.addSystemClass(gaeClass);
```
aozarov: Ah, that answer my question before (regarding *always favoring one version*).
What would happen if we didn't do that (which I hope would allow a simpler include/exclude set)? Will it *always* favor the version bundled with the application (even for jetty/mvm-runtime) usage? If so, would you consider that as a problem?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  public void configure(WebAppContext context) throws Exception {
+    ClassLoader loader = context.getClassLoader();
+    try {
+      // Test if the appengine api is available
+      loader.loadClass(GAE_CHECK_CLASS);
+    } catch (Exception ex) {
+      VmRuntimeWebAppContext.logger.log(Level.FINE, "No appengined API", ex);
```
aozarov: Nit, would that be nicer?

```java
if (VmRuntimeWebAppContext.logger.isLoggable(Level.FINE)) {
  VmRuntimeWebAppContext.logger.log(Level.WARNING, "No appengine API jar included in WEB-INF/lib! Please update your SDK!", ex);
} else {
  VmRuntimeWebAppContext.logger.log(Level.WARNING, "No appengine API jar included in WEB-INF/lib! Please update your SDK!");
}
```

Regardless, I am not sure how useful the exception is going to be in any case.
And if you prefer it the way it is then `s/appengined/appengine/` and `s/including/included/`
```

File: jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java [hunk 14]
```diff
@@ -195,6 +187,28 @@ protected void doStart() throws Exception {
     }
 
     super.doStart();
+
+    // Look for a datastore service within the context
+    ClassLoader orig = Thread.currentThread().getContextClassLoader();
```
aozarov: Is that, reflection, to access AE APIs (such as datastore) really necessary (as oppose to specifying AE API jar as "provided" in maven)?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -195,6 +187,28 @@ protected void doStart() throws Exception {
...
+    try {
+      ClassLoader loader = getClassLoader();
+      Thread.currentThread().setContextClassLoader(loader);
+      Class<?> factory = loader.loadClass(DatastoreServiceFactory.class.getName());
+      contextDatastoreService = factory.getMethod("getDatastoreService").invoke(null);
+      if (contextDatastoreService != null) {
+        getActiveTransactions =
+            contextDatastoreService.getClass().getMethod("getActiveTransactions");
+        getActiveTransactions.setAccessible(true);
```
aozarov: is that needed? its a public method.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -504,25 +505,40 @@ private void complete(Request baseRequest) {
...
-              "Request completed without committing or rolling back transaction with id "
-                  + txn.getId()
-                  + ".  Transaction will be rolled back.");
-          txn.rollback();
-        } catch (Exception e) {
-          // We swallow exceptions so that there is no risk of our cleanup
-          // impacting the actual result of the request.
+          Object txns = getActiveTransactions.invoke(contextDatastoreService);
+
+          if (txns instanceof Collection) {
```
aozarov: that would always be `true` (return empty collection for `null`) and can be casted upon assignment.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -504,25 +505,40 @@ private void complete(Request baseRequest) {
...
+            for (Object tx : (Collection<Object>) txns) {
+              Object id = transactionGetId.invoke(tx);
+              try {
+                logger.warning(
+                    "Request completed without committing or rolling back transaction "
+                        + id
+                        + ".  Transaction will be rolled back.");
+                transactionRollback.invoke(tx);
```
aozarov: could have used `rollbackAsync` for avoid sequential rollbacks.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
```diff
@@ -504,25 +505,40 @@ private void complete(Request baseRequest) {
...
+              } catch (InvocationTargetException ex) {
+                logger.log(
+                    Level.WARNING,
+                    "Swallowing an target exception we received while trying to rollback "
```
aozarov: "a target exception"? Though I would just have said "Failed to rollback abandoned transaction " + id

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeJettyKitchenSinkIT.java
```diff
@@ -63,6 +62,27 @@ public void testJspNotCompiled() throws Exception {
...
   /**
+   * Test that non compiled jstl JSP  can be served.
+   */
+  public void testJstlJSP() throws Exception {
+    String[] lines = fetchUrl(createUrl("/jstl.jsp"));
+    int max = -1;
+    int count = 0;
+    for (String line : lines) {
+      line = line.trim();
+      if (line.length() > 0 && Character.isDigit(line.charAt(0))) {
```
aozarov: You could use `!list.isEmpty()` if you want.

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeJettyKitchenSinkIT.java
```diff
@@ -63,6 +62,27 @@ public void testJspNotCompiled() throws Exception {
...
+        int value = Integer.valueOf(line);
```
aozarov: `parseInt`?

aozarov:
Looks good. Few questions and comments.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
joakime: Some other `javax` packages that come up in the general webapp world (outside of the restricted compat layer)

 * `javax.websocket` (WebSocket stuff, server & client - server side only)
 * `javax.el` (JSTL classes & features - server side & client side)
 * `javax.annotation` (General annotations used by JNDI, CDI, and others, very much server side)
 * `javax.faces` (Java Server Faces - can be server side)
 * `javax.inject` (CDI goodness - can be server side)
 * `javax.ws.rs` (JAXRS stuff - not typically seen server side)
 * `javax.xml.bind` (not typically seen server side)
 * `javax.xml.rpc` (not typically seen server side)
 * `javax.xml.ws` (not typically seen server side)
 * `javax.xml.soap` (not typically seen server side)

gregw: Currently I'm handling all the shared classes the same: they are excluded from the serverClasses (so they are not hidden from the webapp) and they are included in the systemClasses (so they are protected from being overridden by the webapp).

In terms of the different reasons packages are shared, I have just done that by grouping in the source.

Now there are certainly packages in the shared list that need not be protected from replacement (ie they don't need to be system classes), so for example `javax.xml.` could probably be safely replaced by the webapp.   But then there are definitely others that must be protected, like `javax.el` which is shared between the webapp and the JSP implementation.

So without a clear policy or mechanism to determine should a package be protected or not, I've just applied a single shared white list.

I like white lists as they tend to be less vulnerable to accidental change as dependencies change, but in this particular case it may be simpler to revert back to a black list - ie to explicitly hide only those packages (eg `javax.cache.`) that the container requires but are not JVM provided classes.

The resulting list will definitely be shorter, but will be subject to change as we modify the compat implementation and we would be best to generate it rather than author it.

I think the white list is much better for having an explicitly clear policy on all packages.  Black list will be more flexible but less explicit and require more maintenance.         So I've got no clear preference, but Black list is the more normal container solution.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
gregw: javax. is listed as a system class by default in jetty, so it is normal to not let any container provided API to be overridden by the webapp.  But this doesn't extend to other packages provided by the JVM or put on the system classpath.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
gregw: Reflection is necessary here because we are pushing the webapps to providing their own implementation.  While we do have an implementation on the container classpath, it is for our session manager and is a different instance than will be used by the webapp.    So in order to deal with the transactions of the webapp, we need to either use reflection or move the code to a filter that we can load from the webapp classpath.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
gregw: When I wrote the code I didn't have it, but it had permission problems that the setAccessible fixed.   I didn't research exactly why.  I can try again without and look in more detail what permission is failing and why.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
gregw: I just duplicated the existing code.  Shall I change?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
gregw: I prefer having at least a null check as otherwise we are assuming details of getActiveTransactions.   I prefer a instanceof check as then it gives me both a null check and an assert of the returned type.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
gregw: That was the previous message, but I prefer yours!

gregw:
@aozarov I think there are three ways forward for this PR:

1. Stick with the strict whitelist approach as currently implemented
1. Come up with a less strict whitelist, which some JVM classes able to be replaced (but would need a good rational to follow)
1. Revert to a black list approach.

I can't pick between 1 and 3.   I'm happy to create an alternate pull request with 3 implemented, so we can see what that looks like if it will help us choose.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
aozarov: Is adding filters to the app by the container (without user settings) is something unusual or a standard practice? If the latter then I like it better otherwise feel free to leave as is and a comment explaining why we are doing it, such as the one you provided here, would be great!

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
aozarov: Ok, lets keep the white list for now.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
aozarov: it will *never* be `null` it is part of the contract of the (public) method. I understand your argument but on the flip side, we may "silently fail" to do this cleanup if the code is written in a way that does not match our expectations (not to mention that it makes the code harder to read). Feel free to keep as is if you strongly prefer it.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
aozarov: Ah, no, but creating an issue to fix that later would be nice.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeWebAppContext.java
gregw: Containers adding filters that are loaded from their classpath is not unusual,  but containers adding filters that are loaded from the webapps classpath is.

It is a rather strange behaviour really, the container is saying that the webapp must provide its own Datastore implementation, but then intervenes in that impl to impose a transaction rollback policy.     It is a strange halfway house as I would expect that such a container policy would be appropriate to enforce on a container provided services, but not on a service that the container is insisting the webapp provides for itself.

I better solution would be for the jar that provides the datastore service to also contain a utility filter for enforcing a rollback policy.  Applications could then choose to instantiate it or not, or it would be easier for the container to put the filter on there classpath if so desired.

aozarov: Ok, lets just add a comment explaining why reflection is used here.

aozarov:
I am fine with (1), I agree that is sounds safer. Would be good to understand the consequence of that - hence a class is provided by both the application *and* container/jre that is not marked as shared. I guess application can get some class mismatch exception if a non shared class is used and passed to a shared class?

aozarov:
@gregw looks good to me!

I think there are few comments which were not addressed (or maybe you prefered not to take any action). In any case, please create a separate issue for the async rollback and anything else that you think we would like to address in the future and merge at will.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/AppengineApiConfiguration.java
gregw: Oops I left the old log line in... stand by...

PR #292, opened 2016-06-22
Ignore comments between XML tags and add XML parsing tests.

ajkannan:
Also removes a [near] duplicated method (`getBody`) in favor of `getText` in `XmlUtils`.

Note that `getText` will concatenate the text content of children nodes, whereas getBody did not.  As this method is only called by code that seeks the text of a node with no (non-comment) children, this shouldn't be a problem.

### DIFF
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/utils/config/AppEngineWebXmlProcessor.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/utils/config/AppEngineWebXmlProcessor.java
@@ -591,7 +591,7 @@ class AppEngineWebXmlProcessor {
     appEngineWebXml.setApiConfig(new ApiConfig(servlet, url));
 
     for (Element subNode : getNodeIterable(node, "endpoint-servlet-mapping-id")) {
-      String id = trim(XmlUtils.getText(subNode));
+      String id = XmlUtils.getText(subNode);
       if (id != null && id.length() > 0) {
         appEngineWebXml.addApiEndpoint(id);
       }
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/utils/config/AppEngineWebXmlProcessor.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/utils/config/AppEngineWebXmlProcessor.java
@@ -231,48 +231,48 @@ class AppEngineWebXmlProcessor {
   }
 
   private void processApplicationNode(Element node, AppEngineWebXml appEngineWebXml) {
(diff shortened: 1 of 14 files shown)


appengine-managed-runtime/src/main/java/com/google/apphosting/utils/config/AppEngineWebXmlProcessor.java
```diff
@@ -631,14 +631,6 @@ private boolean toBoolean(String value) {
     return (value.equalsIgnoreCase("true") || value.equals("1"));
   }
 
-  private String getTextNode(Element node) {
```
aozarov: This is a bit different than `XmlUtils.getText` as the latter is calling `getTextContent` which will also include the child nodes content (not sure if that is desired or would never happen). If that is fine lets add it to the change description.

appengine-managed-runtime/src/main/java/com/google/apphosting/utils/config/AppEngineWebXmlProcessor.java
```diff
@@ -591,7 +591,7 @@ private void processApiConfigNode(Element node, AppEngineWebXml appEngineWebXml)
     appEngineWebXml.setApiConfig(new ApiConfig(servlet, url));
 
     for (Element subNode : getNodeIterable(node, "endpoint-servlet-mapping-id")) {
-      String id = trim(getTextNode(subNode));
+      String id = trim(XmlUtils.getText(subNode));
```
aozarov: Nit, doesn't `XmlUtils.getText` already trim?

appengine-managed-runtime/src/test/java/com/google/apphosting/utils/config/AppEngineWebXmlProcessorTest.java
```diff
@@ -0,0 +1,2076 @@
...
+    }
+  }
+
+  public void testHealthCheck_getUnhealthyThreshold_negative() {
+    String xml =
+        "<appengine-web-app xmlns=\"http://appengine.google.com/ns/1.0\">\n"
+            + "<application>helloworld-java</application>\n"
+            + "<version>1</version>\n"
+            + "<health-check>\n"
+
```
aozarov: Fine as is but `assertEquals(true, ..` could be replaced with `assertTrue` and `assertEquals(null, ...` could be replaced with `assertNull`

aozarov:
Looks good. I didn't look too carefully at the tests as I assume they were migrated as is (let me know if that is not the case). Few minor suggestions though feel free to merge without it if you want.

Issue #231, opened 2016-05-05
Make Jetty and Jetty compat do log rotations


ludoch:
We do not want to fill the VM disk space with logs.
The environment providing the local logs to the console supports log rotation.
Just need to verify we do it.

gregw:
The request log is setup in jetty9 by [gae.xml](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/async-support/jetty9-base/src/main/jetty-base/etc/gae.xml#L42-L52) and in jetty9-compat by a similar [gae.xml](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/blob/async-support/jetty9-compat-base/src/main/jetty-base/etc/gae.xml#L54-L64)

This indicates we are rolling the logs daily and that we keep the logs for 2 days.

Note that the pure jetty request log does not have a size based log rotation, so we could still fill up the disk.   However, we have the ability to plugin something like [logback](http://logback.qos.ch/access.html) which I believe does have this capability.   @joakime can you confirm that and perhaps provide a PR to change the configuration?    @ludoch please say if you want size based rotation

ludoch:
I think 2 days is too long (there is another agent looking at the /var/log/appengine_logs area and it can catch up to 5 or 10 minutes with rotation..
Logback: can it be configure by time *and* size, i.e rotate every 30mins or when the size is bigger than xMb? That would be ideal.

gregw:
@joakime also it would be good to filter out the routine health check requests

joakime:
With logback, you can do this .

See http://logback.qos.ch/manual/appenders.html#RollingFileAppender

There's 2 ways to do this ...

**Using separate rolling & triggering policies:**

What you'll setup is a `rollingPolicy` which is based on time.
and a `triggeringPolicy` based on file size.

Something like this ...

```xml
<configuration>
  <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
    <file>gae.log</file>

    <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
      <!-- daily rollover -->
      <fileNamePattern>gae.%d{yyyy-MM-dd}.log</fileNamePattern>

      <!-- keep 30 days' worth of history capped at 3GB total size -->
      <maxHistory>30</maxHistory>
      <totalSizeCap>3GB</totalSizeCap>
    </rollingPolicy>

[...]

PR #295, opened 2016-06-30
Issue #231 - Logging Rotation and Levels

joakime:
This PR introduces a bunch of updates to the logging to enable log rotation for both the app*.json and request*.log

We can tweak the rotation numbers in a later commit, but its currently setup as ...

 * app-#.json is at 3 total files of 100MB each (max size)
 * request-#.log is at 3 total files of 10MB each (max size)

This also includes the ability to capture log4j & commons-logging & slf4j and route it to the java.util.logging setup.

request logging is also using this new logging layer (not using the older Jetty FileRollover implementation)


appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
```diff
@@ -0,0 +1,124 @@
...
+   * </ol>
+   * </p>
+   *
+   * @param appConfigFile the filename for the (optional) app specific java.util.logging
+   *                      properties file configuration.
+   */
+  public static void init(File appConfigFile) throws IOException {
+    // Use App (User) Configuration specified as a file parameter
+    if (appConfigFile != null && appConfigFile.exists()) {
+      debug("Loading User Config (from file): %s", appConfigFile);
```
aozarov: Do you think the context would be enough to make "User Config" clear or should we say "User log config..." ?

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
```diff
@@ -0,0 +1,124 @@
...
+      appConfig(appConfigFile);
+    } else {
+      // Use App (User) Configuration specified as a System property
+      String julConfigFile = System.getProperty(JAVA_UTIL_LOGGING_CONFIG_PROPERTY);
+      if (julConfigFile != null) {
+        File configFile = new File(julConfigFile);
+        if (configFile.exists()) {
+          debug("Loading User Config (from property): %s", appConfigFile);
```
aozarov: shouldn't appConfigFile be `configFile` here and bellow (in the warning)?

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
```diff
@@ -0,0 +1,124 @@
...
+        debug("No User Config");
+      }
+    }
+
+    // Manually Adjust Configuration to support System Logging Requirements
+    systemConfig();
+  }
+
+  /**
+   * Convenience method for {@link #init(File)}
```
aozarov: Nit, missing '.'

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * @param appConfigFilename the filename of the config file (or null)
+   * @throws IOException if unable to configure the logging
+   */
+  public static void init(String appConfigFilename) throws IOException {
+    File appConfigFile = null;
```
aozarov: Is this method used? Personally, I prefer to avoid "convenience" methods unless used or requested. regardless I think code could be simplified to:
```java
init(appConfigFileName == null ? null : new File(appConfigFileName));
```

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
```diff
@@ -0,0 +1,124 @@
...
+    if (appConfigFilename != null) {
+      appConfigFile = new File(appConfigFilename);
+    }
+    init(appConfigFile);
+  }
+
+  private static void appConfig(File configFile) {
+    try (FileInputStream is = new FileInputStream(configFile)) {
+      LogManager logManager = LogManager.getLogManager();
+      logManager.reset(); // close & remove existing handlers, reset existing logger levels
```
aozarov: I think this is done as part of `readConfiguration(is)`

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
```diff
@@ -0,0 +1,124 @@
...
+    // that need to be initialized during this step in the logging initialization.
+
+    int count = 0;
+    ServiceLoader<SystemLogger> serviceLoader = ServiceLoader.load(SystemLogger.class);
+    for (SystemLogger systemLogger : serviceLoader) {
+      debug("Executing SystemLogger: %s", systemLogger.getClass().getName());
+      systemLogger.configure();
+      count++;
+    }
+    debug("Ran %d SystemLogger(s)", count);
```
aozarov: Ran -> Initialized | Configured ?

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
```diff
@@ -0,0 +1,124 @@
...
+    init(appConfigFile);
+  }
+
+  private static void appConfig(File configFile) {
+    try (FileInputStream is = new FileInputStream(configFile)) {
+      LogManager logManager = LogManager.getLogManager();
+      logManager.reset(); // close & remove existing handlers, reset existing logger levels
+      logManager.readConfiguration(is);
+    } catch (SecurityException | IOException e) {
+      System.err.println("Warning: caught exception when reading logging properties: " + configFile
```
aozarov: `warning(...` ?

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
```diff
@@ -0,0 +1,37 @@
...
+  public void run() {
+    ThreadLocalRandom rand = ThreadLocalRandom.current();
+    LOG.trace(String.format("A CommonsLogging Trace Event: %d", rand.nextInt()));
+    LOG.debug(String.format("A CommonsLogging Debug Event: %d", rand.nextInt()));
+    LOG.info(String.format("A CommonsLogging Info Event: %d", rand.nextInt()));
+    LOG.warn(String.format("A CommonsLogging Warn Event: %d", rand.nextInt()));
+    LOG.error(String.format("A CommonsLogging Error Event: %d", rand.nextInt()),
+        new RuntimeException("Generic Error"));
+  }
+}
```
aozarov: Nit, newline.

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.google.apphosting.logging;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+public class CommonsLoggingExample implements Runnable {
+  private static final Log LOG = LogFactory.getLog(CommonsLoggingExample.class);
```
aozarov: s/LOG/log/

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.google.apphosting.logging;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+public class CommonsLoggingExample implements Runnable {
```
aozarov: does it need to be public?

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+
+public class JsonCaptureHandler extends Handler {
+
+  private JsonFormatter formatter;
+  private List<String> events;
+
+  public JsonCaptureHandler() {
+    super();
```
aozarov: redundant.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package com.google.apphosting.logging;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+
+public class JsonCaptureHandler extends Handler {
```
aozarov: Even though this is part of "test" as they are not test classes consider leaving it with package scope when possible.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private JsonFormatter formatter;
```
aozarov: could be final.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
```diff
@@ -0,0 +1,51 @@
...
+  public void publish(LogRecord record) {
+    events.add(formatter.format(record));
+  }
+
+  @Override
+  public void flush() {
+  }
+
+  @Override
+  public void close() throws SecurityException {
```
aozarov: does not really throw anything.

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
joakime: Well, we could call it "App Config" if that makes it easier to understand.

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
joakime: It should be clear that the configuration file is for the app, not the system or environment.  
And that any changes can be overridden by the system configuration. (documentation?)

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
joakime: Nit, A missing "." should then be added to the checkstyle config**.**  :-)

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
joakime: String is used in runtime, File is used in testing.

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
joakime: I think you are thinking of `autoConfig()`, `reset()` does not occur in `readConfiguration(is)`

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
joakime: Fair enough, changing to 'Initialized'...

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
joakime: ah yah, i added `warning(` later, didn't catch that one.

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
joakime: Nit, newline is not part of the Checkstyle rules. :-)
Also, I cannot find this requirement in https://google.github.io/styleguide/javaguide.html

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
joakime: I'm following style guide here.
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
joakime: Not really, but its just a testcase.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
joakime: will remove

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
joakime: will fix

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
joakime: This follows the signature of the `Handler.close()` interface

joakime: See http://docs.oracle.com/javase/8/docs/api/java/util/logging/Handler.html#close--

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
aozarov: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/logging/LogManager.java#749

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
```diff
@@ -0,0 +1,124 @@
...
+    init(appConfigFile);
+  }
+
+  private static void appConfig(File configFile) {
+    try (FileInputStream is = new FileInputStream(configFile)) {
+      LogManager logManager = LogManager.getLogManager();
+      logManager.reset(); // close & remove existing handlers, reset existing logger levels
+      logManager.readConfiguration(is);
+    } catch (SecurityException | IOException e) {
+      warning("Warning: caught exception when reading logging properties: %s", configFile
```
aozarov: Is the "Warning" prefix needed now?

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
aozarov: > Nit, newline is not part of the Checkstyle rules. :-)

Indeed. That is not part of the google guide line but it is a common practice and most tools complain about it (including github code review tool). Do you mind to update our checkstyle to include `<module name="NewlineAtEndOfFile">` ?

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
aozarov: Yes, but as subclass you don't have to specify it if you are not actually throwing it.

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
aozarov: what I mean is that don't you need to pass `configFile` to the debug method?
Also, in the case of the warning bellow you can get NPE if `appConfigFile` is `null`.

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
aozarov: Right, some of the javadoc related checks were disabled initially as we have too many violations. We should probably create an issue to enable them and at the same time fix the issues.

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
aozarov: Fine as is, though because this is used by tests but not really a test file I would be more strict about it.

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
aozarov: From that exact segment:
"Every constant is a static final field, but not all static final fields are constants. Before choosing constant case, consider whether the field really feels like a constant. For example, if any of that instance's observable state can change, it is almost certainly not a constant."

Also, see the example for that section regarding the logger.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
```diff
@@ -85,19 +104,155 @@ public void messageIncludesStackTrace() throws Exception {
...
+
+  @Test
+  public void testCommonsLogging() {
+    final Instant now = Instant.now();
+    final String expectedThreadName = Thread.currentThread().getName();
+
+    // Generate Events from Commons Logging
+    new CommonsLoggingExample().run();
+
+    // Verify that event was captured by slf4j -> java.util.logging
```
aozarov: maybe the expected should be returned by `CommonsLoggingExample().run()`?

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
```diff
@@ -85,19 +104,155 @@ public void messageIncludesStackTrace() throws Exception {
...
+    String[][] expected = new String[][] {{"DEBUG", "A CommonsLogging Debug Event: "},
+        {"INFO", "A CommonsLogging Info Event: "}, {"WARNING", "A CommonsLogging Warn Event: "},
+        {"ERROR", "A CommonsLogging Error Event: "}};
+
+    List<String> events = capturingHandler.getEvents();
```
aozarov: could this verification be done in a common method?

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
```diff
@@ -71,34 +69,18 @@ private static String fileLogPattern() {
...
-   * Initialize the {@code VmRuntimeFileLogHandler} by installing it on the root logger.
-   */
-  public static void init() throws IOException {
-    reloadLoggingProperties(LogManager.getLogManager());
+  @Override
+  public void configure() throws IOException {
     Logger rootLogger = Logger.getLogger("");
+
+    // Look for the expected Handlers
     for (Handler handler : rootLogger.getHandlers()) {
```
aozarov: Fine as is. A java 8 alternative (which I am not convinced is more readable) could be:
```java
Optional<Handler> handler =
    rootLogger.getHandlers().filter(x -> x instance of VmRuntimeFileLogHandler).findAny();
if (!handler.isPresent()) {
  rootLogger.addHandler(new VmRuntimeFileLogHandler());
}

File: appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java [hunk 19]
```diff
@@ -31,22 +31,20 @@
...
   public static final String LOG_DIRECTORY_PROPERTY = "com.google.apphosting.logs";
   public static final String LOG_PATTERN_CONFIG_PROPERTY =
       "com.google.apphosting.vmruntime.VmRuntimeFileLogHandler.pattern";
   // Log files to /var/log/app_engine/app.[0-2].log.json
   private static final String DEFAULT_LOG_DIRECTORY = "/var/log/app_engine";
-  private static final String DEFAULT_LOG_PATTERN = "app.%g.log.json";
+  private static final String DEFAULT_LOG_PATTERN = "app.%u.%g.log.json";
   private static final String APP_ENGINE_LOG_CONFIG_PATTERN_ENV = "APP_ENGINE_LOG_CONFIG_PATTERN";
-  private static final int LOG_MAX_SIZE = 100 * 1024 * 1024;
+  private static final int LOG_PER_FILE_SIZE = 100 * 1024 * 1024;
```
aozarov: Here and in `RequestLoggerHandler` is there a way to make that configurable?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/RequestLoggerHandler.java
```diff
@@ -0,0 +1,69 @@
...
+
+    String pattern = System.getProperty(LOG_PATTERN_CONFIG_PROPERTY);
+    if (pattern != null) {
+      if (pattern.startsWith("/")) {
+        return pattern;
+      }
+      return directory + "/" + pattern;
+    }
+
+    return directory + "/" + DEFAULT_LOG_PATTERN;
```
aozarov: Any reason not to return DEFAULT_LOG_PATTERN as the default value for its associated `System.getProperty` as you do for directory?

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/JettyRunner.java
```diff
@@ -255,11 +267,61 @@ public void run() {
...
+   * attempt to find the specific pattern you are interested in.
+   * </p>
+   *
+   * @param regex the text (regex) pattern to look for in the log file.
+   * @return the File for the discovered log file
+   * @throws java.io.FileNotFoundException if unable to find a log
+   * file with the specific text pattern
+   * @throws IOException if unable to read a log file and/or pattern
+   */
+  public File findLogFileThatContains(String regex) throws IOException {
```
aozarov: can we move this function to the test?

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
joakime: j.u.l logging Handlers should generally be public, and accessible from j.u.l as it can (and does) restart Handlers and/or move them to deeper loggers on its own terms.

aozarov: Sounds like a good reason to keep as is then.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
joakime: I want to capture the fact that each logging impl has different levels that map to a j.u.l level.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
joakime: Was wanting to keep the example as pristine as possible, let the testcase assert the expected results instead.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
joakime: What is the preferred configuration technique?

  System properties?  (probably best choice)
  System environment variables?  (some things in your library does this, and I disagree with this approach)
  config file? (don't agree with this one either)

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/RequestLoggerHandler.java
joakime: Just following the configuration techniques that were there before in the VmRuntimeFileLogHandler (sp?)

If these configuration values are not being used prior to this effort (ie: backward compatibility), then I'll be happy to make these configuration values easier to setup.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
aozarov: Then lets make it configured via system property for now.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/RequestLoggerHandler.java
aozarov: I think I only suggest code style change without behaviour change:
```java
String pattern = System.getProperty(LOG_PATTERN_CONFIG_PROPERTY, DEFAULT_LOG_PATTERN);
if (pattern.startsWith("/")) {
  return pattern;
}
String directory = System.getProperty(LOG_DIRECTORY_PROPERTY, DEFAULT_LOG_DIRECTORY);
return directory + "/" + pattern;
```

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
gregw: @joakime isn't there a way that we can get arbitary parameters out of the logging.properties file.   I think we briefly discussed this so that you could set something like `java.util.logging.FileHandler.fileSize=12345`

gregw:
@joakime with @aozarov away, I will review/test these changes before we ask @ifigotin to review/merge.  So once you have updated for the latest comments and merged to match the latest async-support, ping me and I'll run some tests.

ludoch:
@joakime: can you describe more how this helps #318?

gregw:
@ludoch I don't think this issue helps/hinders #318, my testing shows same (bad) results for with and without this PR.   This PR is simply to establish the log rotation for both logs and a bit of an implementation cleanup.    #318 will required changing what we put into the logs.

ludoch:
@aozarov Would you have some cycles to review this PR? Thanks.

ludoch:
ack

aslo:
@aozarov Do you have a chance to take another look at this PR? Thanks!

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
ludoch: @joakime to reply

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
ludoch: Is it done?

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
ludoch: What is the status?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/RequestLoggerHandler.java
ludoch: Can we close on this one?

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/JettyRunner.java
ludoch: Can it be done?

joakime: This is a test class already.  not sure what you are asking.

appengine-java-logging/src/main/java/com/google/apphosting/logging/CoreLogging.java
joakime: Corrected in commit 846e4d2223c26b1cd9752e7e907c1d4ed47301c8

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
joakime: Fixed the code (0c55fd766dd76fe7d6a4385fc88bc2e3b512cce9), but not the checkstyle rules (should probably be a different PR).

Why do we want to have different rules then the google style guide?

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
joakime: Status is, not implemented.

The following properties are supported now (this is default behavior of java.util.logging):

 *  `com.google.apphosting.vmruntime.VmRuntimeFileLogHandler.limit` is the `LOG_PER_FILE_SIZE`
 * `com.google.apphosting.vmruntime.VmRuntimeFileLogHandler.count` is the `LOG_MAX_FILES`

But the problem is, you cannot default those values to anything meaningful, they HAVE to be defined for those values to be used (otherwise they result in no-value, or default behavior).  You cannot reset those values after the construction of the Handler too, they either have to be passed into the Handler constructor *or* be pulled from the properties.  There's no other way.

It also makes no sense for this specific VmRuntimeFileLogHandler to allow property based configuration, as this isn't configured from the WebApp side, its a System Logger that doesn't use the properties files to initialize.

This logger is initialized from ServiceLoader, at startup of Jetty, with some options configurable via system properties and runtime environment variables.

Also, the webapp side logging has no access to even initialize this VmRuntimeFileLogHandler (due to the ClassLoader isolation) anyway, meaning you couldn't set this up and configure it as part of the webapp side logging.properties file.

The only meaningful way to configure this is outside of the webapp, at the server / system level.
Do we want to support that?   I would discourage that, as you would then have hierarchies of property files (which is often confusing for people).

Example scenario:

 * Cloud logging.properties - initialized VmRuntimeFileLogHandler (so that fluent can access the logs)
 * Jetty's logging.properties - initializes the jetty's slf4j-api -> slf4j-jdk14 layer, and adds the RequestLoggerHandler (this file could also have jetty specific logging levels and possibly extra handlers)
 * Compat level, webapp logging.properties - configures the webapp's specific logging requirements.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/RequestLoggerHandler.java
joakime: Yes, as the proposed alternative approach breaks the backward compatibility with the `LOG_PATTERN_CONFIG_PROPERTY` configuration option

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
aozarov: It is not a contradicting rule but rather a less "strict"/"enforced" rule. 
The "internal" version of the google style guide also mention of being consistent with the current file and the project. I believe most files in this project comply with this rule, and if so, this one should too.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+
+public class JsonCaptureHandler extends Handler {
+  private final JsonFormatter formatter;
+  private List<String> events;
+
+  public JsonCaptureHandler() {
+    formatter = new JsonFormatter();
```
aozarov: Nit. Any reason for not assigning in the declaration and avoiding the need for an explicit constructor? Also, could events be final ?

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
aozarov: still not addressed. Do you plan to subclass this class? If so, then make sense if not then please change (and probably make the class final).

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonCaptureHandler.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public void flush() {
+  }
+
+  @Override
+  public void close() throws SecurityException {
+  }
+
+  public List<String> getEvents() {
+    return events;
```
aozarov: Should this return an immutable view of the list?

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
```diff
@@ -85,19 +104,155 @@ public void messageIncludesStackTrace() throws Exception {
...
-        reader
-            .readLine()
-            .startsWith("\tat " + getClass().getName() + ".messageIncludesStackTrace"));
+    assertTrue(reader.readLine()
+        .startsWith("\tat " + getClass().getName() + ".messageIncludesStackTrace"));
+  }
+
+  @Test
+  public void testCommonsLogging() {
+    final Instant now = Instant.now();
```
aozarov: avoid final for params and local variables

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
```diff
@@ -85,19 +104,155 @@ public void messageIncludesStackTrace() throws Exception {
...
+    String[][] expected = new String[][] {{"DEBUG", "A CommonsLogging Debug Event: "},
+        {"INFO", "A CommonsLogging Info Event: "}, {"WARNING", "A CommonsLogging Warn Event: "},
+        {"ERROR", "A CommonsLogging Error Event: "}};
+
+    List<String> events = capturingHandler.getEvents();
+    assertThat("Events", events.size(), is(expected.length));
+
+    for (int i = 0; i < events.size(); i++) {
+      String logLine = events.get(i);
+      JsonData data = new Gson().fromJson(logLine, JsonData.class);
```
aozarov: Fine as is (for a test), but the same `Gson` instance could have been used for all events.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
aozarov: That would be clear enough with the params (expected[][] and probably classname for logger message match) that you pass to the common method from all 3 testing methods.

appengine-java-logging/src/test/java/com/google/apphosting/logging/CommonsLoggingExample.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.google.apphosting.logging;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+class CommonsLoggingExample implements Runnable {
+  private static final Log LOG = LogFactory.getLog(CommonsLoggingExample.class);
```
aozarov: Use lowercase, here and in the other logger deceleration in this PR (not a "constant").
See [this](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names).

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/RequestLoggerHandler.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.io.IOException;
+import java.util.logging.FileHandler;
+import java.util.logging.Formatter;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+
+public class RequestLoggerHandler extends FileHandler implements SystemLogger {
```
aozarov: javadoc

aozarov:
Added some comments. Didn't finish my pass. Should be done with it tomorrow.

joakime:
Wow @aozarov 

Thanks for the review, I'll cover these current ones tomm with a few extra commits to this PR.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
aozarov: I think that we should, at the very least, be able to use a System property to override the constant value.

joakime: Sure thing, easy enough to do.

I propose we use the same names as java.util.logging would have used.

Config | Property Key
-------- | ---------
LOG_PER_FILE_SIZE | com.google.apphosting.vmruntime.VmRuntimeFileLogHandler.limit
LOG_MAX_FILES | com.google.apphosting.vmruntime.VmRuntimeFileLogHandler.count

aozarov: Also, a side question. Where would [ServletContext.log](http://docs.oracle.com/javaee/6/api/javax/servlet/ServletContext.html#log(java.lang.String)) go?

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/RequestLoggerHandler.java
```diff
@@ -0,0 +1,69 @@
...
+  public static class RequestLogFormatter extends Formatter {
+    @Override
+    public String format(LogRecord record) {
+      // We don't expand the message, or use the ResourceBundle stuff for this formatter.
+      return record.getMessage();
+    }
+  }
+
```
aozarov: remove empty line

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/JettyRunner.java
aozarov: This class is not a test but rather a test helper.
Currently (I think) the only test that is actually using this method is `LoggingIT` which is where I was suggesting to move it to.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
```diff
@@ -32,22 +32,20 @@
  * with the cloud logging agent.
  *
  */
-public class VmRuntimeFileLogHandler extends FileHandler {
-
+public class VmRuntimeFileLogHandler extends FileHandler implements SystemLogger {
```
aozarov: I wonder if these 2 handlers (this and `RequestLoggerHandler` - app/access) should not be located at `appengine-java-logging`. The reason is, I think they are also relevant to vanilla jetty and maybe should be configured in that level?

aozarov:
Done with my review.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
joakime: Each logger has different levels as well, by the time you implement it this way you are back to juts as verbose as above.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
joakime: ServletContext.log() goes to the Jetty's ContextHandler.

The logger name is either the webapp context's display-name, or the context-path.

In the configuration for this project, the logger goes to jetty -> slf4j-api -> slf4j-jdk14 -> these system loggers

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmRuntimeFileLogHandler.java
joakime: They could be, but I was leaving them in-place for backward compatibility reasons.

joakime: On rethink ...

The RequestLogHandler is really 100% jetty specific, and not useful outside of that context.

jetty9-compat-base/src/main/java/com/google/apphosting/vmruntime/jetty9/RequestLoggerHandler.java
```diff
@@ -10,6 +10,12 @@
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 
+/**
+ * The {@link java.util.logging.Handler} responsible for capturing the
```
aozarov: Nit, "is responsible"

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/LoggingIT.java
```diff
@@ -28,11 +28,63 @@
...
+   * layer will generate a new log file each time its reset/initialized
+   * because it sees the lock from the previous instance before the
+   * previous instance has a chance to close and remove the lock.
+   * </p>
+   * <p>
+   * This function will iterate through all of the log files and
+   * attempt to find the specific pattern you are interested in.
+   * </p>
+   *
+   * @param regex the text (regex) pattern to look for in the log file.
```
aozarov: Nit, missing param for `logDir`. Also, param should not end with '.'

jetty9-compat-base/src/test/java/com/google/apphosting/vmruntime/jetty9/LoggingIT.java
```diff
@@ -28,11 +28,63 @@
...
+   * @return the File for the discovered log file
+   * @throws java.io.FileNotFoundException if unable to find a log
+   * file with the specific text pattern
```
aozarov: fix wrapping.

appengine-java-logging/src/test/java/com/google/apphosting/logging/JsonFormatterTest.java
aozarov: I am not sure what you mean by "Each logger has different levels as well". I am not suggesting to eliminate the different `expected` arrays but rather have these tests:
```
assertThat("severity", data.severity, is(expected[i][0]));
assertThat("timestamp.seconds", data.timestamp.seconds, ...
assertThat("timestamp.nanos", data.timestamp.nanos,
assertThat("thread name", data.thread, is(expectedThreadName));
assertThat("logger message", data.message, ..
if (data.severity.equals("ERROR")) {
  ...
}
```

which seems to be the same factored into one method that will receive `expected` and `loggerMessagePrefix` as params.

aozarov:
Looks good. I think [this](https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/pull/295/files#r76836294) comment should still be addressed but fine with merging it as is.

ludoch:
LGTM, will merge soon.

ludoch:
Changes in gcr.io/google_appengine/jetty9-compat:githubheadasync

ludoch:
revert is pushed as 

gcr.io/google_appengine/jetty9-compat:2016-09-02_00000

and 

gcr.io/google_appengine/jetty9-compat:githubheadasync

Issue #302, opened 2016-07-23
 Java/jetty kills the app 5 seconds after SIGTERM


ludoch:
A customer is is having problems with the Java runtime stopping the process 5 seconds after we send SIGTERM. The lifecycle PRD specifies that apps should have 30 seconds at least.
Our Jetty.xml has
```
<Set name="stopTimeout">5000</Set>
```

ludoch:
I think we should do the change in the 2 images (master and async)

gregw:
Annoyingly 5000 is hard coded in jetty.xml, which makes it difficult to change.  I have opened https://github.com/eclipse/jetty.project/issues/774 to fix this, but will prepare a PR to work around it for now.

ludoch:
We also need a fix in master

ludoch:
I guess, we would also need that in the pure jetty9 runtimes (master and async?)

gregw:
@ludoch Commit ef87bbaa86b31609059c6f1576f5b69fd30a8a65 did apply the longer time to the pure jetty9 runtime in async.       While there is a gke-jetty in master, it doesn't really have a base directory like we do in async, so a different approach will be needed.   Is that image used?

ludoch:
Someone told me this:
"ah crap, signals are not forwarded in by the jetty_run.sh script
but it seems doable though
http://unix.stackexchange.com/questions/146756/forward-sigterm-to-child-in-bash"
???

gregw:
OK, I will update in both master and async; pure and compat.   Shall I use this issue and PR?

ludoch:
sure.

ludoch:
I guess it can be tested via sshing to the docker container?

gregw:
@ludoch that's how I'm testing... standby...

PR #305, opened 2016-07-27
Configurable stop time #302

gregw:
Temporary fix for #302 until parameter added to jetty release

### DIFF
--- a/jetty9-compat-base/src/main/jetty-base/modules/gae.mod
+++ b/jetty9-compat-base/src/main/jetty-base/modules/gae.mod
@@ -39,6 +39,7 @@ jetty.httpConfig.requestHeaderSize=8192
 jetty.httpConfig.responseHeaderSize=8192
 jetty.httpConfig.sendServerVersion=true
 jetty.httpConfig.sendDateHeader=false
+jetty.server.stopTimeout=30000
 jetty.server.dumpAfterStart=false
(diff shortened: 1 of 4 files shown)


jetty9-compat-base/src/main/jetty-base/modules/gae.mod
```diff
@@ -39,6 +39,7 @@ jetty.httpConfig.requestHeaderSize=8192
 jetty.httpConfig.responseHeaderSize=8192
 jetty.httpConfig.sendServerVersion=true
 jetty.httpConfig.sendDateHeader=false
+jetty.server.stopTimeout=30000
```
ludoch: Not commented here, but commented on the one above... jetty9-base/src/main/jetty-base/modules/gae.mod
Any reasons?

gregw: There is a default value in the XML of 30s that will be used if the property is not set.

The property is commented out in the jetty9 module because it is following the jetty convention of just documenting with a commented property how to change the default

The property is not commented out in the compat module because it is following the gae-compat convention of being explicit in one place of key values and not relying on jetty defaults.

ludoch: ok

ludoch:
LGTM

Issue #306, opened 2016-07-28
Hiding app.yaml at web app root


ajkannan:
It would be good to avoid serving the app.yaml in the staged app's root directory.  Perhaps we could add a snippet to webdefault.xml?

janbartel:
We could use jetty-specific methods of hiding app.yaml. Using security constraints in webdefault.xml would be difficult:  if using quickstart, then the webdefault.xml comes from the sdk and we have no control over it. If not using quickstart then we can of course modify the webdefault.xml.

I would instead suggest that we use the o.e.j.webapp.WebAppContext's support for hiding arbitrary resources. See [here](https://github.com/eclipse/jetty.project/blob/jetty-9.3.x/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/WebAppContext.java#L98)  and [here](https://github.com/eclipse/jetty.project/blob/jetty-9.3.x/jetty-webapp/src/main/java/org/eclipse/jetty/webapp/WebAppContext.java#L269).  

For the standard jetty image, we can use our DeploymentManager to add a protected target of "/app.yaml" to every deployed webapp.

For compat, (as it does not use the DeploymentManager) we can instead hard-code the exclusion of "/app.yaml" in our WebAppContext subclass.

Of course, it might be best to ensure that the app.yaml is not included in the deployable, if at all possible.

PR #307, opened 2016-07-28
Hide app.yaml #306

gregw:
Hide app.yaml #306

For the pure jetty image, a deployment binding is used to apply a `gae-web.xml` (name needs to be changed when other `gae` names are changed) file to every context deployed.

For compat, the protectedTarget is just directly set.

### DIFF
--- a/jetty9-base/pom.xml
+++ b/jetty9-base/pom.xml
@@ -122,7 +122,7 @@
           <arguments>
             <argument>-jar</argument>
             <argument>../jetty-distribution-${jetty.version}/start.jar</argument>
-            <argument>--add-to-startd=gae,http,deploy,jsp,jstl</argument>
+            <argument>--add-to-startd=gae,http,jsp,jstl</argument>
           </arguments>
         </configuration>
       </plugin>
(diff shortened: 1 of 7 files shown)


ludoch:
Just realized that for a pure jetty9 deployment, we are deploying a packaged jar file and not an exploded war. This WAR file might not contain the app.yaml (at least via the Maven plugin -I did not try Gradle) as this file is treated as a Maven resource... Need to verify though.

gregw:
For the pur jetty9 deployment, we are using the normal deployer, which will actually allow multiple webapps to be deployed, which obviously will not be supported by the SDK.  But who is to say how app images will eventually be aggregated and deployed.

So it is well worth the effort to make a deployment binding to have the gae-web.xml file as a place to apply arbitrary policy to any deployed webapp, such as protecting app.yaml.   This is a very low cost, even if it turns out that app.yaml is never there.       Thus I think it is a good thing to add and if it eventually proves needless, we can remove as an optimization.

ludoch:
Sounds good.

ludoch:
We also need this to the master branch, still used by customers.

PR #310, opened 2016-08-04
Issue #302 configurable stop timeout

gregw:
Issue #302 configurable stop timeout


gregw:
@ludoch update to exec in master.

PR #311, opened 2016-08-04
Issue #302, exec jetty_cmd.sh so that signals are passed to JVM

gregw:
Issue #302, exec jetty_cmd.sh so that signals are passed to JVM


ludoch:
Tested via ssh to the vm?

gregw:
@ludoch I ran the container with the bash command and then manually ran the jetty-run.sh command and verified that the java command was exec and took over the pid of the original script.  The following worked:
```bash
cd $JETTY_BASE
./jetty_run.sh &
kill -3 %!
```
The signal is delivered to JVM and a thread dump is correctly generated.

Issue #313, opened 2016-08-04
"Could not create logging file: No such file or directory"


meltsufin:
Steps to reproduce:

1) build the project
2) $ docker run openjdk8:async-8-jre
```
...
Could not create logging file: No such file or directory
COULD NOT CREATE A LOGGINGFILE 20160804-202320.1!openjdk version "1.8.0_91"
OpenJDK Runtime Environment (build 1.8.0_91-8u91-b14-1~bpo8+1-b14)
OpenJDK 64-Bit Server VM (build 25.91-b14, mixed mode)
...
```

gregw:
@meltsufin pull request #315 is quick fix for this.  It creates the log directory, but the name needs to be reviewed.

meltsufin:
Thanks, Greg! Should we create a separate issue to review the naming?

PR #315, opened 2016-08-04
Create /var/log/app_engine directory #313

gregw:
Create /var/log/app_engine directory #313


meltsufin:
LGTM

ludoch:
What is the side effect on Compat, (or even flex) where I think this directory path is provided as a Docker volume? Someone has to check this PR on a real image for both pure openjdk and compat...

meltsufin:
That's a good point. The issue was raised based on a local run of the image without mounting the logs volume, but if the volume is always mounted when the App Engine run the image, it may not be an actual issue. We just need to fix the README to instruct mounting of the logs volume.

gregw:
no longer required

PR #317, opened 2016-08-09
Change Maven coordinates of the logging module

meltsufin:
This disconnects the module from the parent pom, and changes the Maven coordinates to:

**groupId:** com.google.cloud.runtimes
**artifactId:** logging
**version:** 0.1.0-SNAPSHOT

PTAL @aslo @gregw @ludoch


aslo:
LGTM.

PR #321, opened 2016-08-11
Disconnect individual repos from parent pom

meltsufin:
I copied the parent pom into each of the individual repos that we're planning.
I also changed the Maven coordinates for openjdk-runtime and jetty-runtime modules.
Each one of the repos should build now without the top level parent pom.

@aslo @gregw @ludoch PTAL


aslo:
Took a look at this, and LGTM. Each module builds locally so I'm satisfied. Should we also go ahead and delete the root-level parent pom in this PR as well?

One other thing I just want to point out is that once these modules are fully split into different repos, the build for the `jetty-runtime` will be broken until the logging module (with the new maven coordinates) is available from maven central, since it's no longer a submodule of the same project. So, we'll probably want to do a maven release of that lib immediately once the repos are migrated so as to not block development.

meltsufin:
Thanks! I just kept the parent pom for now for easy building and for the CI to work.

PR #324, opened 2016-08-18
Move READMEs to the root of the future repos

meltsufin:
I moved the READMEs to the roots of the repos, but I didn't really fix them up yet. That will be a separate PR, probably after the repos are created.

@aslo PTAL


aslo:
LGTM

Issue #322, opened 2016-08-16
Use GAE_MEMORY_MB for heap size calculation


meltsufin:
Currently /proc/meminfo MemTotal is used, but GAE_MEMORY_MB is the new environment variable that has a more accurate number.

gregw:
I've not yet seen that in the images I've deployed.  For a 4GB instance will that variable be `4096` or `4096M`.

We can change the current setup-env.bash to look for this and only fall back to meminfo if not found, or it may be simpler to just change after the restructure?

rdcastro:
No suffix, only 4096.

meltsufin:
@gregw I think we need to change setup-env.bash so that we use GAE_MEMORY_MB if available, and fall back to /proc/meminfo MemTotal otherwise. The reason being that GKE is not going to have GAE_MEMORY_MB.

meltsufin:
@rdcastro GAE_MEMORY_MB is now available in production for `vm: true` and `env: flex`, but I'm not sure what the new heapsize formula should be.
It used to be:
**heap_size = host_total_memory * 0.8 - 400**.
Should the new formula be
**heap_size = GAE_MEMORY_MB * 0.8**?

rdcastro:
It seems your previous formula was "80% of what is available to the app container". GAE_MEMORY_MB is now the amount available to the app container (which is also enforced via Docker), so I think you can keep the same formula as before if it worked well.

In case it's not documented, please make sure that is documented somewhere so users can understand that if they requested 1Gb, they will only have 800Mb for the Java app, which might not be obvious.

meltsufin:
@joakime do you have some cycles to look at this?

joakime:
The code change has already been made locally.
Time permitting it will have a PR submitted. 
But know that the full cycle for testing even small changes like this can take many hours.

meltsufin:
Great, thanks!

gregw:
replaced by https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/3

PR #325, opened 2016-08-26
Issue #322 - Use new GAE_MEMORY_MB variable if provided

joakime:
+ The HEAP_SIZE configuration in setup-env.bash will now
  favor the ${GAE_MEMORY_MB} if present, using the value
  as-is, with no percentage adjustments, and then fall
  back to the old 80% of /proc/meminfo routines we
  had before

### DIFF
--- a/openjdk8/src/main/docker/setup-env.bash
+++ b/openjdk8/src/main/docker/setup-env.bash
@@ -21,7 +21,7 @@ fi
 
 # Setup default Java Options
 : ${JAVA_TMP_OPTS:=$( if [[ -z ${TMPDIR} ]]; then echo ""; else echo "-Djava.io.tmpdir=$TMPDIR"; fi)}
-: ${HEAP_SIZE:=$(awk -v frac=${HEAP_SIZE_FRAC:=0.8} -v res=${RAM_RESERVED_MB:=400} /MemTotal/'{ print int($2/1024*frac-res) "M" } ' /proc/meminfo)}
+: ${HEAP_SIZE:=$(if [[ -z ${GAE_MEMORY_MB} ]]; then awk -v frac=${HEAP_SIZE_FRAC:=0.8} -v res=${RAM_RESERVED_MB:=400} /MemTotal/'{ print int($2/1024*frac-res) "M" } ' /proc/meminfo ; else echo "${GAE_MEMORY_MB}M"; fi)}
 : ${JAVA_HEAP_OPTS:=-Xms${HEAP_SIZE} -Xmx${HEAP_SIZE}}
 : ${JAVA_GC_OPTS:=-XX:+PrintCommandLineFlags -XX:+UseG1GC -XX:+PrintGCDateStamps -XX:+PrintGCDetails -XX:+ParallelRefProcEnabled}
 : ${JAVA_GC_LOG_OPTS:=$( if [[ -z ${JAVA_GC_LOG} ]]; then echo ""; else echo "-Xloggc:${JAVA_GC_LOG} -XX:+UseGCLogFileRotation -XX:GCLogFileSize=1048576 -XX:NumberOfGCLogFiles=4"; fi)}


joakime:
Tested on `vm:true` and `vm:env-flex`.

Results:

```
MemTest Servlet

Total Memory: 1,396,703,232 b
 Free Memory: 1,243,955,320 b
 Used Memory:   152,747,912 b
  Max Memory: 1,396,703,232 b
env[GAE_MEMORY_MB] = 1331
```

meltsufin:
I think we still want to take 80% of GAE_MEMORY_MB for the heap size. The only thing we're not doing to GAE_MEMORY_MB is subtracting 400MB of overhead memory.

@aozarov Is it something you can confirm?

joakime:
@meltsufin i thought the whole point of GAE_MEMORY_MB was that it was a different value then `/proc/meminfo` intentionally meant to convey the memory available to the GAE process itself (regardless if its Java or Python or Go, etc)

If its going to be 80% of GAE_MEMORY_MB, then what's the point of GAE_MEMORY_MB?

meltsufin:
@joakime I'm also trying to understand where that 80% rule came from originally. My understanding is that the problem with `/proc/meminfo` is that it's reporting host VM memory, not the memory allocated to the Docker container. So, before GAE_MEMORY_MB, we were compensating for it by subtracting 400MB from `/proc/meminfo`. However, we were also only assigning 80% of the container available memory to Java heap. I'm not sure what the exact rationale was for that number, but it's important to note that we can't dedicate all of the available memory to the Java heap. There's also some non-heap memory that the JVM needs. I believe we can control the non-heap memory with `XX:MaxPermSize`, where the default is 64MB. Hence, I think 80% leaves too much memory unused for large VMs.
Maybe Java heap should be `GAE_MEMORY_MB - MaxPermSize`.
@ludoch @aozarov What do you think?

meltsufin:
Also, I just realized that `MaxPermSize` has been replaced in Java 8 with `MaxMetaspaceSize`. In addition, we have to consider the stack memory used by each thread, which can be controlled with JVM arg `-Xss`. Since we don't know how many threads an app will use, maybe the 80% of GAE_MEMORY_MB is actually not a bad default.

joakime:
With java 8, and a desire for a good default configuration of our runtime, would recommend not setting `MaxMetaspaceSize` (the `MaxPermSize` equivalent in Java 8) setting.

Using that is just a sign of bad premature optimization.
Let the GC handle it, it will do a great job without that tweak.

If a user has a need for a more custom configuration of the GC, they'll need to weak the `setup-env.sh` anyway.

joakime:
So this is the result of the current implementation.

```
MemTest Servlet

Total Memory: 1,396,703,232 b
 Free Memory: 1,234,806,960 b
 Used Memory:   161,896,272 b
  Max Memory: 1,396,703,232 b
env[GAE_MEMORY_MB] = 1331
env[HEAP_SIZE] = null

dump of /proc/meminfo
MemTotal:        1745344 kB
MemFree:          293440 kB
MemAvailable:     831888 kB
Buffers:           36016 kB
Cached:           591608 kB
SwapCached:            0 kB
Active:           902508 kB
Inactive:         437380 kB
Active(anon):     715492 kB
Inactive(anon):      208 kB
Active(file):     187016 kB
Inactive(file):   437172 kB
Unevictable:        4076 kB
Mlocked:            4076 kB
SwapTotal:             0 kB
SwapFree:              0 kB
Dirty:                44 kB
Writeback:             0 kB
AnonPages:        716336 kB
Mapped:           123368 kB
Shmem:               980 kB
Slab:              72040 kB
SReclaimable:      53768 kB
SUnreclaim:        18272 kB
KernelStack:        3616 kB
PageTables:         6380 kB
NFS_Unstable:          0 kB
Bounce:                0 kB
WritebackTmp:          0 kB
CommitLimit:      872672 kB
Committed_AS:    2519240 kB
VmallocTotal:   34359738367 kB
VmallocUsed:        5792 kB
VmallocChunk:   34359725420 kB
HardwareCorrupted:     0 kB
AnonHugePages:         0 kB
HugePages_Total:       0
HugePages_Free:        0
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB
DirectMap4k:       40948 kB
DirectMap2M:     1740800 kB
DirectMap1G:           0 kB
```

It appears that `GAE_MEMORY_MB` is already 80% of `/proc/meminfo/MemTotal`, and the environment even has 300MB still free after the 1,331MB pre-allocation that the JVM has.

meltsufin:
`GAE_MEMORY_MB` is supposed to be the amount of memory provided to the App Engine runtime Docker container. It's host memory minus 0.4GB.

meltsufin:
According to [Oracle GC Ergonomics](http://docs.oracle.com/javase/8/docs/technotes/guides/vm/gc-ergonomics.html):
**initial heap size** = Larger of 1/64th of the machine's physical memory on the machine or some reasonable minimum. 
**maximum heap size** = Smaller of 1/4th of the physical memory or 1GB.

In our case, these numbers would be based on the total physical memory of the VM, not GAE_MEMORY_MB. It does seem like a lot of available memory would by unutilized if we stick with the JVM defaults.

In any case, the users can always pass in the JVM startup arguments through app.yaml.
Something like:
```
env_variables:
  JAVA_HEAP_OPTS: -Xms500m -Xmx500m
```

gregw:
Seeing as we know the JVM is the primary process in each image, leaving 3/4 of the available memory unused does not feel correct.

Thus I'm still in favour of setting the heap size by default (and we need to add the -server flag as well)

meltsufin:
Agreed. I favor sticking with the 80% of the GAE_MEMORY_MB as the default.

gregw:
replaced by https://github.com/GoogleCloudPlatform/openjdk-runtime/issues/3

PR #327, opened 2016-09-02
Revert "Issue #231 - Logging Rotation and Levels"

ludoch:
Reverts GoogleCloudPlatform/appengine-java-vm-runtime#295


ludoch:
revert is pushed as 

gcr.io/google_appengine/jetty9-compat:2016-09-02_00000

and 

gcr.io/google_appengine/jetty9-compat:githubheadasync

PR #329, opened 2016-09-28
README: Docker for Mac troubleshooting

hraban:
Maven expects to find docker at localhost:2375, but on Docker for Mac
it's at /var/run/docker.sock.


hraban:
This caused me a lot of head scratching; didn't realise 2375 was docker itself and not some expected mock GAE service.

PR #330, opened 2016-09-28
Switch base image from qa debian to debian8

sharifelgamal:
cc @dlorenc


meltsufin:
LGTM

meltsufin:
@sharifelgamal We need a similar change here https://github.com/GoogleCloudPlatform/openjdk-runtime

PR #341, opened 2016-12-09
Jetty http client async support

meltsufin:
Re-creating the PR from a local branch.
I added a small fix to the pom to make the image include jetty-client.jar.
@olamy

### DIFF
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
@@ -111,9 +111,7 @@ public class VmApiProxyDelegate implements ApiProxy.Delegate<VmApiProxyEnvironme
   private static HttpClient createHttpClient( long connectTimeout) {
     try {
       HttpClientTransportOverHTTP transport = new HttpClientTransportOverHTTP();
-      // FIXME what sort of customize SslContextFactory we want to use here??
-      SslContextFactory sslContextFactory = new SslContextFactory();
-      HttpClient httpClient = new HttpClient( transport, sslContextFactory );
+      HttpClient httpClient = new HttpClient( transport, null);
       httpClient.setMaxConnectionsPerDestination( VmApiProxyEnvironment.MAX_CONCURRENT_API_CALLS );
       httpClient.setConnectTimeout( connectTimeout + ADDITIONAL_HTTP_TIMEOUT_BUFFER_MS );
       httpClient.setTCPNoDelay( true );
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
@@ -38,33 +38,33 @@ import com.google.apphosting.api.UserServicePb.CreateLoginURLResponse;
 import com.google.apphosting.api.UserServicePb.CreateLogoutURLResponse;
 import com.google.apphosting.utils.remoteapi.RemoteApiPb;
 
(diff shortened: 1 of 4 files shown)


appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
```diff
@@ -98,29 +97,46 @@
...
+  public VmApiProxyDelegate() {
+    this(createHttpClient());
   }
 
-  public VmApiProxyDelegate() {
-    this(new DefaultHttpClient(createConnectionManager()));
+  private static HttpClient createHttpClient() {
+    try {
+      HttpClientTransportOverHTTP transport = new HttpClientTransportOverHTTP();
+      // FIXME what sort of customize SslContextFactory we want to use here??
```
meltsufin: It doesn't look like any requests go over SSL. So, we should be ok with not using one at all.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
```diff
@@ -98,29 +97,46 @@
...
+      SslContextFactory sslContextFactory = new SslContextFactory();
+      HttpClient httpClient = new HttpClient( transport, sslContextFactory );
+      httpClient.setMaxConnectionsPerDestination( VmApiProxyEnvironment.MAX_CONCURRENT_API_CALLS );
+      httpClient.setConnectTimeout( ADDITIONAL_HTTP_TIMEOUT_BUFFER_MS );
```
meltsufin: Originally it was `timeoutMs + ADDITIONAL_HTTP_TIMEOUT_BUFFER_MS`. I know that `timeoutMs` is not available at this point, but maybe there is a workaround?

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
```diff
@@ -360,48 +381,34 @@ RuntimeException constructApiException(String packageName, String methodName) {
...
-        CoreConnectionPNames.SO_TIMEOUT, timeoutMs + ADDITIONAL_HTTP_TIMEOUT_BUFFER_MS);
+    request.header(RPC_STUB_ID_HEADER, REQUEST_STUB_ID);
+    request.header(RPC_METHOD_HEADER, REQUEST_STUB_METHOD);
 
-    // Performance tweaks.
-    params.setBooleanParameter(CoreConnectionPNames.TCP_NODELAY, Boolean.TRUE);
-    params.setBooleanParameter(CoreConnectionPNames.STALE_CONNECTION_CHECK, Boolean.FALSE);
-    request.setParams(params);
+
+    request.timeout( ADDITIONAL_HTTP_TIMEOUT_BUFFER_MS, TimeUnit.MILLISECONDS );
```
meltsufin: missing `timeoutMs +`?

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
olamy: correct we only have to remove the FIXME :-)

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/VmApiProxyDelegate.java
olamy: @meltsufin AFAICS the connection timeout was alway set to DEFAULT_RPC_TIMEOUT_MS. So I will add to have similar behaviour.

meltsufin:
LGTM

PR #342, opened 2016-12-10
Async support merge

gregw:
This is a merge of master to the async-support branch to keep the branch point up to date.

Most of the changes were not appropriate, but I believe these ones should be merged over.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,11 @@
+[![deprecated](http://badges.github.io/stability-badges/dist/deprecated.svg)](http://github.com/badges/stability-badges)
+
+We've moved!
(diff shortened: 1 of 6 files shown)


meltsufin:
Yep, these are good! LGTM

gregw:
@meltsufin my git-foo has failed me, as merging branches via a PR has not actually moved the branch point of the real branches.     I think we now need to do a real merge from master to async-support, ignore all the changes and then commit?

meltsufin:
I've merged in the changes in this PR to async-support. Are you saying you that not all changes that you intended to include were included here?

PR #343, opened 2016-12-12
add unit test using embedded Jetty to test real http call

olamy:
Signed-off-by: olivier lamy <olamy@apache.org>


ludoch:
Nice. 
In general, all comments (doc or inlined comments) should start with a capital letter and finish with a period.

PR #347, opened 2017-02-01
[WIP] Add support for logging based on google-cloud-logging

meltsufin:
I'm trying to port over the new logging mechanism to the compat runtime to be used with a custom `logging.properties` file, but not having much luck.

In my test app, I have this in `webapp\WEB-INF\logging.properties`:
```
# Set the default logging level for all loggers to INFO
.level=INFO

# Override root level
com.foo.level=FINE

# Override parent's level
com.foo.bar.level=SEVERE

handlers=com.google.cloud.logging.LoggingHandler
com.google.cloud.logging.LoggingHandler.level=FINE
com.google.cloud.logging.LoggingHandler.log=gae_app.log
com.google.cloud.logging.LoggingHandler.resourceType=gae_app
com.google.cloud.logging.LoggingHandler.enhancers=com.google.cloud.logging.GaeFlexLoggingEnhancer
com.google.cloud.logging.LoggingHandler.formatter=java.util.logging.SimpleFormatter
java.util.logging.SimpleFormatter.format=%3$s: %5$s%6$s
```

and `appengine-web.xml` has this:
```XML
<appengine-web-app xmlns="http://appengine.google.com/ns/1.0">
  <application>YOUR-PROJECT-ID</application>
  <version>YOUR-VERSION-ID</version>
  <threadsafe>true</threadsafe>
  <env>flex</env>
  <beta-settings>
    <setting name="machine_type" value="n1-standard-1"/>
    <setting name="enable_app_engine_apis" value="true"/>
  </beta-settings>
  <system-properties>
    <property name="java.util.logging.config.file" value="WEB-INF/logging.properties"/>
  </system-properties>
</appengine-web-app>
```

@gregw This configuration doesn't seem to be working. I only see logs in the `app` stream and no sign of  `gae_app.log` in the Stackdriver Logging UI. Any thoughts?

### DIFF
--- a/appengine-managed-runtime/pom.xml
+++ b/appengine-managed-runtime/pom.xml
@@ -31,6 +31,7 @@
   <properties>
     <maven.build.timestamp.format>yyyyMMddHHmm</maven.build.timestamp.format>
     <appengine.api.version>1.9.40</appengine.api.version>
+    <gcloud.api.version>0.8.2-beta-SNAPSHOT</gcloud.api.version>
   </properties>
 
   <dependencies>
@@ -86,7 +87,7 @@
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-server</artifactId>
       <version>${jetty.version}</version>
-      <scope>test</scope>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
@@ -95,6 +96,17 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-logging</artifactId>
+      <version>${gcloud.api.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.servlet</groupId>
+          <artifactId>servlet-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
   </dependencies>
 
   <build>
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/RequestContextScope.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/RequestContextScope.java
@@ -1,0 +1,82 @@
+package com.google.apphosting.vmruntime;/*
+ * Copyright (C) 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


appengine-managed-runtime/pom.xml
```diff
@@ -31,6 +31,7 @@
   <properties>
     <maven.build.timestamp.format>yyyyMMddHHmm</maven.build.timestamp.format>
     <appengine.api.version>1.9.40</appengine.api.version>
```
lesv: This is probably no longer used, but if it is, it should be at least 1.9.48 if not .49

appengine-managed-runtime/pom.xml
```diff
@@ -31,6 +31,7 @@
...
+    <gcloud.api.version>0.8.2-beta-SNAPSHOT</gcloud.api.version>
```
lesv: Why are you using the -SNAPSHOT ??

You do need to initialize Cloud Logging and turn on the JUL hook to use it. -- I don't see that here.

gregw:
@meltsufin I'm not going to have a chance to run this myself for a few hours.  But I'm guessing the problem is that this compat runs with the working directory at $JETTY_BASE rather than in the webapp itself. So your relative path to `WEB-INF/logging.properties` is not working.   Try putting an absolute path there to confirm/deny

@lesv the SNAPSHOT is used because the `LoggingHandler` from google-cloud-java with the modifications to be enhanced for the monitored resource and traceid has not yet been released.

appengine-managed-runtime/pom.xml
meltsufin: Yep, thanks!

appengine-managed-runtime/pom.xml
meltsufin: Like Greg mentioned, we don't have a non-snapshot release of it yet.

meltsufin:
@gregw It looks like `WEB-INF/logging.properties` is being picked up fine based of the debugging logs that I added. I think the issue has to do with the initialization of the logging handler. It's probably failing, but we don't see any logs for it anywhere.

gregw:
@meltsufin perhaps we need a PR on google-cloud-logging that will print any exceptions to stderr so we one can better debug logging problems in init?

meltsufin:
@gregw Yes, that's a good idea. I also think we need to pull the zone stuff into this PR for now because it sounds like they won't be able to merge that PR and release in a while.

PR #349, opened 2017-03-09
Fix big file handling

gregw:
Set the response buffer size to the max response size.

This PR also updates the debian image to be setup the same as https://github.com/GoogleCloudPlatform/openjdk-runtime, including using openjdk-8.   This was done because the non-durable debian repositories have changed and it is not possible to build the previous image as it was.   Use of java-8 was chosen because java-7 is end-of-life and the source/target of the `maven-compiler-plugin` was already set the 1.8 in the pom!

### DIFF
--- a/appengine-jetty-managed-runtime/src/test/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeJettyKitchenSinkTest.java
+++ b/appengine-jetty-managed-runtime/src/test/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeJettyKitchenSinkTest.java
@@ -20,18 +20,18 @@ import com.google.apphosting.api.ApiProxy;
 import com.google.apphosting.vmruntime.VmApiProxyDelegate;
 import com.google.apphosting.vmruntime.VmApiProxyEnvironment;
 import com.google.apphosting.vmruntime.VmRuntimeUtils;
-import static com.google.apphosting.vmruntime.jetty9.VmRuntimeTestBase.PROJECT;
-import static com.google.apphosting.vmruntime.jetty9.VmRuntimeTestBase.VERSION;
+
+import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
 import java.net.HttpURLConnection;
 import java.net.URL;
 
 import java.util.Arrays;
 import java.util.List;
 import javax.servlet.http.HttpServletResponse;
-import static junit.framework.TestCase.assertEquals;
-import static junit.framework.TestCase.assertNotNull;
-import static junit.framework.TestCase.assertTrue;
 import org.apache.commons.httpclient.Header;
 import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.methods.GetMethod;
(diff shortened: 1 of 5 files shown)


appengine-jetty-managed-runtime/src/test/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeJettyKitchenSinkTest.java
```diff
@@ -42,11 +42,47 @@
  */
 public class VmRuntimeJettyKitchenSinkTest extends VmRuntimeTestBase {
 
-	
-
 	@Override
 	protected void setUp() throws Exception {
 		appengineWebXml = "WEB-INF/sessions-disabled-appengine-web.xml";
+		File webinf = new File("./WEB-INF");
```
meltsufin: Formatting seems off

gke-debian-openjdk/src/main/docker/Dockerfile
```diff
@@ -17,44 +17,25 @@ ENV DEBIAN_FRONTEND noninteractive
...
 # Update debian 
 RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/jessie-backports.list \
- && apt-get -q update \
- && apt-get -y -q --no-install-recommends install \
-    ca-certificates \
-    openjdk-8-jre \
+  && apt-get -q update \
+  && apt-get -y -q --no-install-recommends install \
+    ca-certificates-java=20161107'*' \
+    openjdk-8-jre-headless=8u121'-*' \
```
meltsufin: I don't think you need to specify `8u121` version here. It will pick up latest anyway.

gke-debian-openjdk/src/main/docker/Dockerfile
```diff
@@ -17,44 +17,25 @@ ENV DEBIAN_FRONTEND noninteractive
...
     netbase \
     wget \ 
     unzip \
  && apt-get clean \
  && rm /var/lib/apt/lists/*_*
 
-# workaround for https://bugs.debian.org/775775
```
meltsufin: I would leave all of this old code below, just to minimize the risk of this PR. The only lines we need changed are 22 and 23 above.

meltsufin:
@gregw I made the changes from my review. The fix seems to work. Can you just review my changes to make sure I didn't overlook anything.

I do see `java.io.FileNotFoundException: WEB-INF/logging.properties (No such file or directory)` in the logs, but I guess it was always there when specifying the property in `appengine-web.xml`
```XML
  <system-properties>
    <property name="java.util.logging.config.file" value="WEB-INF/logging.properties"/>
  </system-properties>
```

gregw:
@meltsufin I got confused by bringing in some other fixes that used java 7.  So when I replaced that it looked like I was upgrading.

reviewing changes shortly...

gregw:
I can't approve review my own PR, but your changes are good and pass local/remote tests for me.

meltsufin:
@gregw A customer is reporting much increased memory usage of the app. Do you think the buffer size setting is causing it? Maybe they have many concurrent requests.

joakime:
Is the customer generating large responses? (over 32MB) or small?

meltsufin:
Asked, waiting for a response.

meltsufin:
Does `response.setBufferSize(CommitDelayingOutputStream.MAX_RESPONSE_SIZE_BYTES)` cause a lot of direct memory to be allocated? I'm noticing a huge increase in the JVM memory usage as soon as this line is added.

gregw:
The setBufferSize call does not immediately allocated, but as soon as any writing is done, a buffer of that size will be allocated.    It's obviously a too large buffer.

Note that the `CommitDelaying` wrappers really are not doing what they represent, as they look like they are delaying commits until 32MB, but really only use the default response buffer size of 32KB.

If delaying commit so that logging headers may be added to the response is still a necessity, then we should update the filter to do it's own aggregation into a growing buffer (or list of buffers).    If it is no longer a required feature, then we should look at removing the delaying commit filter as it is mostly a noop.

meltsufin:
I would assume that delaying the commit until the response is completed is still a requirement. I think we should just fix the `CommitDelaying` wrappers.

gregw:
OK, I'll prepare a PR early next week.  

My preference is for creating a chain of standard sized (32KB) buffers taken from the jetty buffer pool, rather than allocating large and larger buffers and doing copies, as there can be many copies on the way from 32KB to 32MB.  It also means than the released buffers are more usable.

gregw:
@meltsufin work in progress at https://github.com/GoogleCloudPlatform/appengine-java-vm-runtime/compare/master-delayCommit

Is currently passing existing tests, but I'd like to take a bit more time to consider if those tests are extensive enough and to test real deployments.

meltsufin:
@gregw We're deprioritizing this work because we got confirmation that #352 fixes the customer's issue with truncated responses without increasing memory usage. So, I'll merge that PR and release the runtime. 

With that being the case, this branch is then more of an enhancement, which is not of high priority given that compat is deprecated. You can leave the branch around, but I wouldn't spend much more time on it at the moment.

gregw:
@meltsufin OK - that's good as while the buffering is simple enough (and looks to be working fine and ready if needed), it really is an inefficient thing to do.

Note however, that even the existing delay committing wrapper (which doesn't really delay the commit) is inefficient as it blocks Jetty from efficiently serving static content from file mapped buffers.   If usage of this image is to continue for any large loads, it is probably worthwhile turning off this filter/wrapper as has been done on more recent images.

PR #352, opened 2017-03-17
Remove overwrite of response buffer size and just warn on committed response

meltsufin:
This PR rolls back the change in #349 which apparently causes very high memory use in the JVM.
Instead, this PR enables handling of large responses by just warning about the possibility of log messages being lost.

You might see messages like these in the logs.

```
com.google.apphosting.vmruntime.jetty9.VmRuntimeWebAppContext doScope: Response for request to '/big.jsp' was already committed (code=200). This might result in lost log messages.' 
com.google.apphosting.vmruntime.VmAppLogsWriter waitForCurrentFlush: End of request or previous flush has not yet completed, blocking.
```

However, the responses should not be truncated.

### DIFF
--- a/appengine-jetty-managed-runtime/src/test/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeJettyKitchenSinkTest.java
+++ b/appengine-jetty-managed-runtime/src/test/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeJettyKitchenSinkTest.java
@@ -23,6 +23,7 @@ import com.google.apphosting.vmruntime.VmRuntimeUtils;
 import org.apache.commons.httpclient.Header;
 import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.methods.GetMethod;
+import org.junit.Ignore;
 
 import javax.servlet.http.HttpServletResponse;
 import java.io.*;
@@ -125,7 +126,7 @@ public class VmRuntimeJettyKitchenSinkTest extends VmRuntimeTestBase {
     assertEquals("4095", lines[4095]);
   }
 
-  public void testHugeTxt() throws Exception {
+  public void ignore_testHugeTxt() throws Exception {
     System.err.println("Expect: java.io.IOException: Max response size exceeded.");
     HttpURLConnection connection = (HttpURLConnection) createUrl("/huge.txt").openConnection();
     connection.connect();
--- a/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/CommitDelayingResponse.java
+++ b/appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/CommitDelayingResponse.java
@@ -77,9 +77,6 @@ public class CommitDelayingResponse extends HttpServletResponseWrapper {
    */
   public CommitDelayingResponse(HttpServletResponse response) throws IOException {
     super(response);
-    // Set the buffer size explicitly, as this ensure both that jetty's aggregate
-    // buffer is the correct size and that the commit threshold is that size.
-    response.setBufferSize(CommitDelayingOutputStream.MAX_RESPONSE_SIZE_BYTES);
     this.output = new CommitDelayingOutputStream(super.getOutputStream());
   }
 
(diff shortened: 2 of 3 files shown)


meltsufin:
Interestingly, now the `VmRuntimeJettyKitchenSinkTest.testHugeTxt()` test is failing. However, previously it was passing for the wrong reason. The expected `java.io.IOException: Max response size exceeded.` was never thrown. It was the `ServletException` that this PR removes instead.

appengine-jetty-managed-runtime/src/test/java/com/google/apphosting/vmruntime/jetty9/VmRuntimeJettyKitchenSinkTest.java
```diff
@@ -125,7 +126,7 @@ public void testBigTxt() throws Exception {
     assertEquals("4095", lines[4095]);
   }
 
-  public void testHugeTxt() throws Exception {
+  public void ignore_testHugeTxt() throws Exception {
```
joakime: Why not just use `@Ignore("Old behavior, now deprecated")` ?

meltsufin: I tried `@Ignore` but it didn't work. Looks like this test is using JUnit 3.

appengine-managed-runtime/src/main/java/com/google/apphosting/vmruntime/CommitDelayingResponse.java
```diff
@@ -77,9 +77,6 @@
    */
   public CommitDelayingResponse(HttpServletResponse response) throws IOException {
     super(response);
-    // Set the buffer size explicitly, as this ensure both that jetty's aggregate
-    // buffer is the correct size and that the commit threshold is that size.
-    response.setBufferSize(CommitDelayingOutputStream.MAX_RESPONSE_SIZE_BYTES);
```
gregw: It may be better to replace this line with:
```
    response.setBufferSize(response.getBufferSize());
```
as this will reset the the commit threshhold (by default 1/4 of the buffer size) to be the buffer size.    By default if the buffer size is 4K, then any individual write larger  than 1K will be handled with a direct write rather than aggregation into the buffer.

Note also that the default buffer size is 32KB, so it may be worth a little memory to increase this to say 128KB to ensure that more resources are committed after generation.     The  prior 32MB size was obviously way too large.

Note that the output stream wrapper is not really doing anything and could pretty much be removed.   It currently will cause a hard write failure at 32MB, but if the response is already committed and being sent over a streaming transport, then there is little reason to have the hard fail.

Alternately, we could re-implement the original GAE implementation that had growing buffers that could grow the 32MB before failure.

meltsufin: Does Jetty use a pool of direct byte buffers for the responses? When we set the buffer to 32MB, how much off-heap memory can it potentially use up?

The hard fail at 32MB is fine because that mimics the behavior of Standard. I do like the idea of modifying the output stream wrapper to hold the response in a growing buffer up to 32MB. I believe that was the original intent, but wasn't correctly implemented. Also, the buffer should be growing and allocated on the heap.

meltsufin: @gregw What is the default commit threshold in Jetty without setting the response buffer size? If it is very small, I feel like we should've been hearing about the truncation problem all the time.

gregw: By default the output buffer is 32KB and the commit write threshhold is 8KB.  Note that this means that if first write is larger than 8KB in a single call to write, then it will be directly written (and a commit done).    But if the initial write is < 8KB then we allocate a 32KB aggregate buffer and will not commit until it is full (or flush is called).

So there are probably few apps that write dynamic content in chunks larger than 8KB.

Jetty does not use direct buffers for aggregate buffers, because they are written too many times and flushed once - hence they are best kept in user space.    We only use direct buffers for caching large content that cannot be file mapped and is written in a single call (and then reused).

We pool heap buffers.

